Issue #13, opened 2015-03-05
Testing Framework


ximex:
We should add an testing framework to check codestyle and a some testcases.
##

ximex:
Automatic Tests with: http://docs.travis-ci.com/user/languages/php/
Testframework: https://phpunit.de/

ximex:
https://codeclimate.com/github/Admidio/admidio

I try some tests. need to exclude the libs folder.
https://github.com/ximex/admidio/commit/a8c2e82c94c27c42d26db0f98127821fa2e0f4c1

ximex:
if we implement the rest api we should create tests for each route

ximex:
i added now the codeclimate.yml file to look how our codequality goes
https://codeclimate.com/github/Admidio/admidio

ximex:
first steps started: aea5b5ca77f7555ce53328fde542f9da66e01bab

kainhofer:
Once the generic CLI utility (PR #2107, issue #2105) is merged, practically all admidio functionality is available via command-line and can be regression tested with a yet-to-be generated unit test suite. Of course, this only covers the core and data handling features, not the pure UI components.

PR #48, opened 2015-05-04
add travis codestyle check

ximex:
first impl for #10 
also needed for #13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: php
+
+php:
+#  - 5.3
+#  - 5.4
+#  - 5.5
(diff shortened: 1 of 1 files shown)


Fasse:
I think it's better to add this to the master and not to the 3.0 branch. There we should only add bugfixes and no "new features".

ximex:
ok then i add it to the master branch.

ximex:
done. and testrun exits with no error.
https://travis-ci.org/ximex/admidio/builds/61376079

To activate it on Admidio/admidio i need admin rights here or you have to register on travis.ci.org and activate the testing

PR #91, opened 2015-07-12
Refactor password

ximex:
please review this code and give feedback


Fasse:
Why do you create a new issue? This is also 89, an improved new hashing method :)

I got the following error when using the branch:
> Parse error: syntax error, unexpected T_OBJECT_OPERATOR in /Users/Jennifer/Sites/GitHub/admidio/adm_program/system/classes/autologin.php on line 119

We should not code so compact as in line 119. This is not easy to read and complicated to debug. Please write more lines and dont combine multiple method calls in one line.

ximex:
What do you mean with "Why do you create a new issue?" This isn't an issue. This is a pull request.
This is what i said in the mails. we could also use pull requests and not only issues ;-)

I think i found the problem: https://stackoverflow.com/questions/13388541/php-parse-error-syntax-error-unexpected-t-object-operator/13388570#13388570

ximex:
b68e7df057480c61cb77d1ac2f7eb20220adddf7 should fix this

Fasse:
Why are here conflicts and how can I solve them?

I dont know how I can see these conflicts. I can to a branch merge to the master but then the Pull-Request is still open ...

ximex:
remove merge from master
rebased on master
force update branch

Issue #55, opened 2015-05-10
Expect MB in preferences where now kB is required


Fasse:
We should change all preferences to MB where we now require kB. Therefore we need to check all places where we use these preferences and calculate kB if required.

ximex:
:+1:

ximex:
i used `KiB`, `MiB`, `GiB`, `TiB` as unit.
look here why i did this:
https://de.wikipedia.org/wiki/Bin%C3%A4rpr%C3%A4fix

PR #106, opened 2015-08-07
change file sizes from kb in mb

ximex:
fix https://github.com/Admidio/admidio/issues/55


ximex:
i leaf the `mib` etc in the code in the method param
https://github.com/Admidio/admidio/commit/60c70f928929a67f1fb1ff4b5f157310eb7cf42a#diff-a7a99ef70cedfad4814b16af1916de7cR179

developers should know the difference between MiB and MB and i think it is important to make that clear

Fasse:
Looks good to me

PR #112, opened 2015-08-08
set hashcost preference by benchmark

ximex:
please check if my logic is correct! didn't tested it ;-)

### DIFF
--- a/adm_program/installation/db_scripts/preferences.php
+++ b/adm_program/installation/db_scripts/preferences.php
@@ -32,6 +32,7 @@ $orga_preferences = array(
     'system_search_similar'         => '1',
     'system_time'                   => 'H:i',
     'system_browser_update_check'   => '0',
+    'system_hashing_cost'           => '10',
 
     // Registration
     'registration_mode'              => '1',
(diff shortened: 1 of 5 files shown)


Fasse:
Please open an issue and describe what you want to do with this code. At the moment I don't know what this is for?!?

Fasse:
Now I have found the time to get familiar with the cost parameter of password_hash. One thing should be considered. The passwort_compat page says that the cost parameter has a range from 4 to 31. We should escape the loop if we reach a cost of 31. And we should begin with cost 4 not with 8 if there is old hardware or something else.

Fasse:
Maybe we should still start with 8 so we do not have to many loops.

ximex:
the loops are not the problem. the first loops would only take less than 20ms.
But i think, that even slow server should have a minimum cost of 8. but i could change this.

Fasse:
The code looks good some minor comments:
- we should declare the globals at the beginning of the method so everyone see which global is already registered and used.
- some code comments would be nice to explain in a short sentence whats going on, especially in update.php and installation.php

ximex:
hope this changes are ok

PR #120, opened 2015-08-19
add option to show browser-update message for old browsers

ximex:


Fasse:
The code seems ok (haven't test the function yet). But the description of the new preference doesn't explain what happen to Admidio if I activate it. It's only a hint that new browsers are better but this is an unnecessary information at this point. We should describe that users of old browsers get a hint to update their browsers.

ximex:
ok i will change that

Fasse:
I have made some checkins to this branch to fix the problem with the preferences. You can merge all of the branch changes to the master. But I could not test the function because I have no old browser :/

ximex:
@Fasse here could you test:
https://www.browserling.com/browse/win/vista/firefox/38/https%3A%2F%2Frfmtc.no-ip.org%2Fsonstiges%2Fadmidio%2Fadm_program%2Findex.php

you have to accept the self signed certificat from my server

PR #129, opened 2015-08-21
admStrTo-Upper/Lower to mb_strto-upper/lower

ximex:
isn't need anymore with PHP min version 5.3.7

https://secure.php.net/manual/en/function.mb-internal-encoding.php
https://secure.php.net/manual/en/function.mb-strtolower.php
https://secure.php.net/manual/en/function.mb-strtoupper.php

### DIFF
--- a/adm_program/installation/installation.php
+++ b/adm_program/installation/installation.php
@@ -374,7 +374,7 @@ elseif($getMode == 6)  // Creating configuration file
                        $gL10n->get('SYS_BACK'), 'layout/back.png');
         }
 
-        $_SESSION['user_email'] = admStrToLower($_SESSION['user_email']);
+        $_SESSION['user_email'] = mb_strtolower($_SESSION['user_email']);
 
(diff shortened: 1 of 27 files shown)


Fasse:
The problem of this function was/is that not every browser support multibyte string. As far as I know multibyte extension is not a php default extension. First we implement the mb_strtolower functions and then there were error reports that this is not working. Therefore we implement the function in *string.php*.

Because of this I would not remove this functions.

ximex:
mehhh... i hate this php extension system...
but maybe we should set mb_internal_encoding

Fasse:
I hate it too. In my opinion there should be 1 php version with a defined set of function/classes ...
What do you mean with maybe we should set mb_internal encoding?

ximex:
set it global: https://github.com/Admidio/admidio/commit/3aa6082c76250c712ae3478fdae585d1da5d1330

Fasse:
But this would return an error if you havn't installed the multibyte extension.

ximex:
oh... yes... you are right...

PR #155, opened 2015-09-26
Use PDO class for database communication

Fasse:
@ximex please review my code


Fasse:
@ximex please have a look at travis. I cant find the resolution to fix this.

ximex:
i think the problems are from the merges of master.
If you want to update a "feature" branch to the changes of master, it is better to do a rebase than a merge.
I will look at it.

Fasse:
No the conflicts are history. I could fix them with the merge. Dont know why this is in the comment.

The Problem with travis is also in the branch since several Versions.

ximex:
found it! look at the missing space
```php
return trim($this->pdo->quote($string),"'");
```
to
```php
return trim($this->pdo->quote($string), "'");
```
pushed some other little improvements

Issue #34, opened 2015-03-30
Use jquery file upload for uploads in download module


Fasse:
Now that we have integrated jquery file upload within the photo module we should also use this plugin for uploads in the download manager.

ximex:
merged

PR #165, opened 2015-10-04
Feature downloads improvements

### LINKED ISSUE #33: Use select2 for roles assignment to folders in download module
When we want to assign roles to folders in download module then we should do this with the select2 control.

Fasse:
fix #33 and #34


Fasse:
@ximex please have a quick look and merge

ximex:
@Fasse to fix issues like the wrong added lines use the possibility to edit commits: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
So we get a nicer commit history.
And git rebase instead of merging ;-)

Issue #203, opened 2015-10-31
Implement date range filter in lists


thomas-rcv:
It should be possible to look for roles in the past and also in the future.
We could use the date range filter like we have in the date module.

See forum:

http://www.admidio.org/forum/viewtopic.php?f=2&t=6692&p=22683#p22683

Fasse:
I had a look to your implementation. The function works as expected but I'm not very happy with the user interface. Within the lists list you don't know what the dates for. Nothing changed if you changed the dates. 
Maybe it's better to place them within the lists presentation and with a better field description so that a user know what changed. Within the event module this was no problem because you know which dates you can filter but here it's a problem. Also with the implementation now you get conflicts with the member status field of mylist. If we have it within the show-list than maybe we can remove the member status field of mylist.
It would also be nice if the filter will only be shown if the role has former or future members.

thomas-rcv:
You are definetly right.
At this time there a several things to do before we could pull it. For example the values of the dates are not set immediatly. Within  the first commit you have to change the category once to get the new date range. I need to add a little bit of javascript for it. 
I didn't tried the behaviour of my lists so far but it doesn't sound good if we already have trouble shooting there. Generally the implementation as I Understood the thread is to have it in list module. My Lists is for my understanding an own module. Sure, both provides list with recordsets at the end but in mylist we have a lot of additional options wirh member states and condition fields.
So if possible I didn't plan to change anything in mylists. Of course if new issues are coming up while the implemantation regarding functional stuff we can discuss and define a strategy how to work on

thomas-rcv:
Please have a look again.
I moved the filter to list_show and also added a text phrase for it. -> we can discuss
I also needed to store some values in the session, using the form.
Some logical improvements and session handling for the variables must be done at this time.
Generelly it should work now, except for minor changes.

PR #212, opened 2015-11-10
reworked #203

thomas-rcv:
-parameters are passed within hidden fields now
- by the way fixed a bug: The list id initialzed with 0 as default by
function admFuncVariableisVald() occures SQL error because lists ids
starts with 1 and are count up.


Fasse:
I found some other things to improve :)
- the new parameters `date_from`and `date_to` aren't documented in the head of lists_show.php.
- In my opinion we should change the navbar description from 

`Mitglieder dieser Gruppe im Zeitraum: Beginn [11.11.2015] Ende [31.12.2025]`

`Filter     Mitglieder im Zeitraum [11.11.2015] bis [11.11.2015]`

Filter is the name of the navbar that we used within the other navbars also. In the other part the text is shorter and the information isn't lost in my opinion. Also within the field date_from and date_to the default value should be only today. That is how the lists worked now. We don't want to show future members within the default. If you call a list you want to see who is now webmaster or in team 1.

thomas-rcv:
Ok this would be minor changes.
How to do this now? Is it possible to change in the current opened pull request or is better to delete it and pull again?

ximex:
you only have to create new commits on the same branch and push it to github. As long as the pullrequest isn't merged it works.

Fasse:
Yes. Just code in your branch and make a commit. The pull request automatically has this new commit

thomas-rcv:
Ok thanks. I guess will have finished tomorrow

Fasse:
I have change the label of the date fields. Please have a look. If it's ok, we can merge.

thomas-rcv:
Is ok for me. Feel Free to merge it

Issue #109, opened 2015-08-08
Refactor dates.php


Fasse:
The following improvements should be done:

- the different print preview modes should be choosen in a new menu form combobox. There should also be the different html (normal and list) modes listed.
- print and html layout should be generated in the same code list, only with different output cases, just like the list module
- print table should use the bootstrap table look and feel, just like the lists print mode
- print preview should use the default layout with headline and table

ximex:
> print table should use the bootstrap table look and feel, just like the lists print mode

should fix partial #59 that more tables got the zebra style

PR #214, opened 2015-11-11
Feature dates list redesign

Fasse:
@thomas-rcv could you du a functional check


thomas-rcv:
Sorry for my late response:

I got some undefined variables for date_view and sometimes for compact-table. In print mode with no matching dates also further notice. The text for option ical export is not matching.
Look at Screens below.
Generelly the implementation is quiet good, but I´m mot really happy with the printer output. A small formatted overview for a month,eg ... , I would fell more confident, like we have so far. We could think about the printers style sheet to improve, or additionally provide Excel export.
One Thing at the end ,... it would be great also outputing  the requested time period in print Output.

mozilla firefox_2015-11-13_23-16-16

Fasse:
You Must make an Update then all notices Are gone.

Fasse:
I have added a notice for the selected period to each print view and also made the print table more condensed. But I don't know what you mean with a small overview of a month or so. This issue was not to add a month view to the module. It only should add bootstrap style to the existing functionallity.

Please have a look at the implementation now. In my view we can merge this.

thomas-rcv:
What I meant is the recordset is printed just on 1 page, like a Excel spreadsheet.
The bootstrap list is very large and you need further pages for printing.
Ok , not part of this issue

thomas-rcv:
I fixed a unkown text key for Ical link.
The implemantation works fine.
From my side push it to master.

Fasse:
#109

Issue #215, opened 2015-11-11
Support "Android String Resources" format for translation files


Fasse:
If we want to manage our translations with Transifex #68 then we should switch the format of our translation files to `Android String Resources` which is supported by Transifex.

ximex:
I think we missed too change the xml files in `demo_data/languages/*.xml`

Fasse:
Ups, you are right.

Issue #68, opened 2015-06-07
Translation with Transifex


ximex:
I see we have already an account there
https://www.transifex.com/organization/Admidio

Fasse:
I have played a little bit with transiflex and it looks like a good solution for translation.
But therefore we must change our xml file. It should be an **Android String Resource**
http://developer.android.com/guide/topics/resources/string-resource.html
This format could be handled by transiflex and is very similar to our current xml.
`<resource>
  <string name="MY_TEXT">My text</string>
</resource>`

ximex:
Here are the possible formats: http://docs.transifex.com/formats/
Maybe use a **json** format instead of a **xml** format?
I'm looking forward to the JSON Rest-API...

Fasse:
Hmm, I uploaded the strings in android format because this was easy to transform. Now I want to try to download in a json format but as I understand the logic of transifex now its not possible to change the format.
Is a JSON format really useful because we still use PHP intern?

ximex:
i think **now** json hasn't any advantage against xml. But if we change it to the Rest API i think it will. But we could leave it yet as xml and change it if we really see that json is better

Fasse:
The code has changed to a transifex format. Now we must use it. Therefore this issue should still be open until all active translators work with it.

Fasse:
Now we have used transifex for english translation and all others know about the new service.

PR #226, opened 2015-11-19
Feature android string resource

Fasse:
#68 #215


Fasse:
@ximex review only needed for the language file classes

thomas-rcv:
I would like to recommend to create a howto in our documentation for Plugin Developers.
Possible language files in Plugins also must be reworked.

Fasse:
do you mean something like this: http://www.admidio.org/dokuwiki/doku.php?id=en:entwickler:how_to_make_your_plugin_translatable
;)

thomas-rcv:
Hmm Looks not so bad. I think we could use it :D

Fasse:
@ximex can i merge? Need this for other changes to language files.

ximex:
checked and improve the code a little bit and rebased. if you are ok with my changes feel free to merge

Fasse:
I have revert the structure of the languages.xml. This is not a translatable file and therefore we don't need the Android Recource String there. Also there were places in the code that were incompatible with the new format.

PR #253, opened 2015-12-08
Refactor admFuncVariableIsValid

ximex:
please check if don't make a mistake.
i will push more commits to remove no longer used variables

### DIFF
--- a/adm_program/installation/installation.php
+++ b/adm_program/installation/installation.php
@@ -137,7 +137,7 @@ elseif(file_exists('../../config.php'))
     exit();
 }
 
-if($getMode === 1)  // (Default) Choose language
+if($getMode === 1) // (Default) Choose language
 {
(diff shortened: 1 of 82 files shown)


thomas-rcv:
Looks ok for me.
Please take into account this function is also used mostly plugins as well.
So eg. deleting 'numeric' would cause a lot of trouble.
If we plan fundamental changes here, we should inform the plugin developer before ddoing so.
No Problem in this pull request

ximex:
I'm not planning to remove numeric ;-) maybe in a far future

Fasse:
You have changed the boolean from 1, 0 to true, false but you dont have changed the calls within the scripts. This may be a problem.

Fasse:
ok, works for me. 1 and 0 are still possible.

PR #256, opened 2015-12-12
Revert "Refactor admFuncVariableIsValid"

Fasse:
Reverts Admidio/admidio#253

Found a problem. Just call the write message script.


Fasse:
Problem will be the `defaultValue = false`

ximex:
maybe this was the problem? https://github.com/Admidio/admidio/commit/7c0b09faec4ed9939e7e28e998fca236a3f1ea53#diff-53729b0d5210d6dc32320bda487de42eL313

Fasse:
https://github.com/Admidio/admidio/blob/master/adm_program/modules/messages/messages_write.php#L35
This fix the problem. But it's not possible to set `defaultValue = false`

If my fix is ok then you can merge it.

Fasse:
@ximex what's about this pull request. Can you have a look at this and my comments.

PR #372, opened 2016-07-20
add SHA512 method for password hashing

ximex:
Missing:
- [ ] Benchmark for SHA512 with rounds option
- [x] Option to set Password-Hashing-Algorithm in config file

### DIFF
--- a/adm_my_files/config_example.php
+++ b/adm_my_files/config_example.php
@@ -47,3 +47,7 @@ $gTimezone = 'Europe/Berlin';
 // for an update of the Admidio database to a new version of Admidio.
 // For a more comfortable and easy update you can set this preference = 0.
 $gLoginForUpdate = 1;
+
+// Set the preferred password hashing algorithm. Default is the PHP default "PASSWORD_DEFAULT"
+// Possible values are: PASSWORD_DEFAULT, PASSWORD_BCRYPT, 'SHA512'
+$gPasswordHashAlgorithm = PASSWORD_DEFAULT;
--- a/adm_program/system/classes/passwordhashing.php
+++ b/adm_program/system/classes/passwordhashing.php
@@ -35,8 +35,8 @@ require_once(SERVER_PATH.'/adm_program/libs/phpass/passwordhash.php');
 class PasswordHashing
 {
     /**
-     * Hash the given password with the given options
-     * Minimum cost is 10
+     * Hash the given password with the given options. The default algorithm uses the password_* methods,
+     * otherwise the builtin helper for SHA-512 crypt hashes from the operating system. Minimum cost is 10.
      * @param string $password  The password string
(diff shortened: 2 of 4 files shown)


adm_program/system/classes/passwordhashing.php
```diff
@@ -69,6 +80,23 @@ public static function verify($password, $hash)
         {
             return password_verify($password, $hash);
         }
+        elseif (strlen($hash) === 98 && strpos($hash, '$6$') === 0)
+        {
+            $passwordHash = crypt($password, substr($hash, 0, 12));
```
ximex: damn i think this substr is wrong... forgot about the rounds...

Fasse:
what's the logic now, will SHA512 automatically used or not?

ximex:
No. The normal password_hashing method is used as default.

adm_my_files/config_example.php
```diff
@@ -47,3 +47,7 @@
 // for an update of the Admidio database to a new version of Admidio.
 // For a more comfortable and easy update you can set this preference = 0.
 $gLoginForUpdate = 1;
+
+// Set the preferred password hashing algorithm. Default is the PHP default "PASSWORD_DEFAULT"
+// Possible values are: PASSWORD_DEFAULT, PASSWORD_BCRYPT, 'SHA512'
```
Fasse: Why don't we use also constant for SHA512? 
This would be more consistent.

ximex: because there is no PHP default constant for SHA512 as int. So i think self defined values should be strings to get no conflict in future.
and you say it: we haven't include our constants here

Fasse: ok, I don't know, that these constants are default php constants.

ximex:
@Fasse the benchmark function is still missing. but i will include it later

Fasse:
@ximex yes I know, but this is nice to have and not neccessary to use the function or am I wrong?

adm_program/system/classes/user.php
```diff
@@ -1213,9 +1213,20 @@ public function setPassword($newPassword, $isNewPassword = false, $doHashing = t
...
+            {
+                $newPasswordHash = PasswordHashing::hash($newPassword, $gPasswordHashAlgorithm, array('rounds' => 5000));
+            }
+            else
+            {
+                $newPasswordHash = PasswordHashing::hash($newPassword, PASSWORD_DEFAULT, array('cost' => $cost));
+            }
 
-            if ($newPassword === false)
+            if ($newPasswordHash === false)
```
wilddom: You should do something with `$newPasswordHash` here, at the moment the plain `$newPassword` gets saved.

ximex: @wilddom thanks. should be fixed b1d9de0ebb5b1bd51873ad41099da1201ee3f006

PR #379, opened 2016-08-18
fix for member assignment

wilddom:


Fasse:
can you please provide a bug description or better reference to a bug issue, so we know what you want to fix. Thanks.

wilddom:
Well, I thought it's pretty obvious. Member assignment is not working. You changed the table to dynamically load data. Doing that is okay, but the eventhandlers of the checkboxes have to be rebound after loading data.

Fasse:
ah ok, now I understand.

Thanks for the fix

Issue #195, opened 2015-10-19
Add relationships between users


Fasse:
We should add relationships between different users. Therefore we need to configure possible relationships and their link to each other. 
Possible relationships roles: 
* mother 
* father
* son
* daughter
* sister
* brother
* ...

Possible links: 
* father - daughter
* father - son
* sister - brother
* sister - sister
* ...

Within the profile you can add relationships to the current user. Therefore you must define his role e.g. **father** and then the user you want to define the relationship and than you can choose the possible relationships roles for this user e.g. **daughter** or **son**. If we consider the gender and this data is mantained in the database we could direkt save the right role e.g. **daughter**.

A link should only defined ones. E.g. the link **father - daughter** could also be **daughter - father** but we should only save one combination. If you choose the role of the current user then we must look for possible links on both assigned roles to links.

ximex:
i think this could be a lot easier.
don't look on the sex (brother/sister). use only "sibling" and "parent" or "ancestor". the sex we get out of the sex property of the user (if defined)
maybe we should also add "boss" or something to add structures in companies etc

Bullnados:
Wohoooo :+1: 
Is the relationship just in same organisation or also between different/more organisations?

ximex:
I don't think that it would get into 3.1
But feel free to implement it @Bullnados ;-)

Bullnados:
I'm in search of somebody which can do that, I can't php. But before starting this big thing, we/you specify the requirements.

thomas-rcv:
Hi all,

I also thought about this issue several times.
I think to have an relationship between members is a specific situation we only have in family memberships, or let´s say I have no have further idea where we could need it.
But in a family role it definitely would make sense.
I´m not sure if it makes sense to delineate the complete structure with son, daughter, etc.

Like Ximex mentioned a parent sibling relationship would also be a good practice for me.
We could think about a hidden role, like we do for participation of events. The leader of the role, or several leaders are the “head” of the role ( Father and Mother )

From sight of the user, or having the right to edit user profiles, I could imaging to have an checkbox in the profile to assign myself as a family Master. Doing like that additional content will fade in for creating/editing further members, like children.
This could be a link to create a new profile. Linking to a new profile the hidden family role with User Id is created first.

In the new profile we would need to think about the profile fields that of course should be optional different. For children we often do not use a own Email address or account information or a login, etc.
Generally this example may only be a problem if the profile fields are defined as mandatory. 
[...]

PR #383, opened 2016-08-27
Feature userrelations

wilddom:
As discussed in #195


adm_program/installation/installation.php
```diff
@@ -685,6 +685,18 @@
                  , ('.$categoryIdSocialNetworks.', \'TEXT\', \'YAHOO_MESSENGER\',\'INS_YAHOO_MESSENGER\', NULL, \'yahoo.png\', NULL, 0, 8, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\') ';
     $db->query($sql);
 
+    // create user relation types
+    $sql = 'INSERT INTO '.TBL_USER_RELATION_TYPES.' (urt_id, urt_name, urt_name_male, urt_name_female, urt_id_inverse, urt_usr_id_create, urt_timestamp_create)
```
Fasse: You must not set the id. If the column is autoincrement then the database will set the id by itself.

wilddom: It's definitely necessary to set the urt_id because it is needed in urt_id_inverse. Just check the following example:

```
urt_id     urt_name     urt_id_inverse
1          parent       2
2          child        1
3          sibling      3
```

adm_program/modules/userrelations/userrelations_new.php
```diff
@@ -0,0 +1,107 @@
...
+$sql = 'SELECT urt_id, urt_name
+          FROM '.TBL_USER_RELATION_TYPES.'
+      ORDER BY urt_name';
+$form->addSelectBoxFromSql('urt_id', $gL10n->get('SYS_USER_RELATION'), $gDb, $sql, array('property' => FIELD_REQUIRED));
+
+// select box showing all users (TODO: check edit right)
+$sql = 'SELECT usr_id, concat(LAST_NAME.usd_value, \' \', FIRST_NAME.usd_value) AS name
+                  FROM '.TBL_MEMBERS.'
+            INNER JOIN '.TBL_ROLES.'
+                    ON rol_id = mem_rol_id
```
Fasse: Add the following code so only users where shown where the current user has the right to view them `rol_id IN (".implode(',', $gCurrentUser->getAllVisibleRoles()).")` 
After that you don't need the join to categories anymore.

adm_program/system/classes/componentupdate.php
```diff
@@ -403,4 +403,22 @@ public function updateStepSetDefaultConfiguration()
...
 
+    /**
+     * This method installs the default user relation types
+     */
+    public function updateStepInstallDefaultUserRelationTypes()
+    {
+        global $gL10n, $gCurrentUser;
+
+        $sql = 'INSERT INTO '.TBL_USER_RELATION_TYPES.' (urt_id, urt_name, urt_name_male, urt_name_female, urt_id_inverse, urt_usr_id_create, urt_timestamp_create)
+            VALUES (1, \''.$gL10n->get('INS_PARENT').'\', \''.$gL10n->get('INS_FATHER').'\', \''.$gL10n->get('INS_MOTHER').'\', 2, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
```
Fasse: Like my comment at installation, you don't need to set the `urt_id`.

wilddom: It's needed because urt_id_inverse is a foreign key relation to urt_id.

Fasse:
👍 
After all a very good job and good code quality. 

Please have a look at my code comments and after you fixed that we can merge this feature.

adm_program/installation/installation.php
Fasse: Ah ok, I don't saw this database relation.

adm_program/modules/userrelations/userrelations_new.php
wilddom: Okay, but when creating a relationship between two users, the current user is required to have the edit right of both users. So, what I actually would need is a list of the users, which can be edited by the currently logged in user. The visible roles are a bit closer but still not perfect... any suggestions?

Fasse: You're right.

So in the first step we should only check `if(!$gCurrentUser->editUsers())` instead of `if(!$gCurrentUser->hasRightEditProfile($user))` like you do within **userrelations_function.php**.

The only users we miss are leaders with the right to edit users but this feature can be added later.

Fasse:
One general hint, you must only edit **de.xml**. 
All other languages will be translated and genereated by transifex.

adm_program/installation/installation.php
```diff
@@ -685,6 +685,18 @@
...
+            VALUES (1, \''.$gL10n->get('INS_PARENT').'\', \''.$gL10n->get('INS_FATHER').'\', \''.$gL10n->get('INS_MOTHER').'\', 2, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
+                 , (2, \''.$gL10n->get('INS_CHILD').'\', \''.$gL10n->get('INS_SON').'\', \''.$gL10n->get('INS_DAUGHTER').'\', 1, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
+                 , (3, \''.$gL10n->get('INS_SIBLING').'\', \''.$gL10n->get('INS_BROTHER').'\', \''.$gL10n->get('INS_SISTER').'\', 3, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
+                 , (4, \''.$gL10n->get('INS_SPOUSE').'\', \''.$gL10n->get('INS_HUSBAND').'\', \''.$gL10n->get('INS_WIFE').'\', 4, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
+                 , (5, \''.$gL10n->get('INS_COHABITANT').'\', \''.$gL10n->get('INS_COHABITANT_MALE').'\', \''.$gL10n->get('INS_COHABITANT_FEMALE').'\', 5, '.$gCurrentUser->getValue('usr_id').',\''. DATETIME_NOW.'\')
```
Fasse: Instead of `$gL10n->get('INS_COHABITANT')` you can use `$gL10n->get('INS_COHABITANT_MALE'). '/'. $gL10n->get('INS_COHABITANT_FEMALE')`, also at other text strings. We don't need the duplicate text strings.

wilddom: Not sure about that. Languages are different, some have gender-neutral/gender-specific terms, others don't. Your example doesn't really work in English: `$gL10n->get('INS_COHABITANT_MALE'). '/'. $gL10n->get('INS_COHABITANT_FEMALE')` would result in cohabitant/cohabitant.

adm_program/modules/userrelations/userrelations_new.php
wilddom: Checking the leader rights should be working now, though not 100% sure it works flawlessly.

adm_program/installation/installation.php
Fasse: But maybe we should use the following syntax:

> Partner/-in:  Heike Mustermann

Within the profile I think this looks better than

> Partner/Partnerin:   Heike Mustermann

Fasse:
Thanks for this great work!

PR #395, opened 2016-09-23
Use select2 when creating new relationships

wilddom:
As requested by @ximex


Fasse:
Why do we need miltiselect if only 1 User could be selected?

wilddom:
We don't, I changed that behaviour in the second commit (d0add2cba2ce80a7f6ec6650a0ecace68e353eee). Before that commit it was not possible to use select2 without a multiselect.

wilddom:
Anything else I should do?

Issue #387, opened 2016-09-03
Typo in lists_default_configuation


ximex:
Fix typo in preferences:
`lists_default_configuation` => `lists_default_configuration`

ximex:
#398

PR #398, opened 2016-10-03
#387 fix typo list_default_configuation

ximex:
@Fasse should the demo data be in a v3.1 "state" or v3.2? => should i also update the demo data?


ximex:
@Fasse hope this is right. feel free to merge

Issue #407, opened 2016-10-06
Extract global vars initialization


ximex:
Extract the initialization of most used global variables into an own file und require that file

ximex:
#408

PR #408, opened 2016-10-06
#407 extract init global in own file

ximex:


ximex:
@Fasse could you review?

Fasse:
Sorry that was not in my view.

PR #420, opened 2016-10-14
Improve logging with PSR-3 logging

### LINKED ISSUE #206: PSR-3 Logging
Implement the logging with the PSR-3 Standard

Standard: http://www.php-fig.org/psr/psr-3/
Lib: https://github.com/Seldaek/monolog
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28014145-psr-3-logging?utm_campaign=plugin&utm_content=tracker%2F10474012&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10474012&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ximex:
fixes #206


Fasse:
@ximex Warum brauchen wir das? Reicht die einfache error_log nicht aus?

Fasse:
The following error occured when install test data: 

> Fatal error: Class 'Monolog\Logger' not found in /Users/Jennifer/Sites/GitHub/admidio/adm_program/system/logging.php on line 18

ximex:
@Fasse hmmm... i think copy something wrong after the last tests... i also delete the test and doc files

Fasse:
Now no error occurs, but there is still no logging in php error file. I have set `$gDebug = 1;`. 

Also admidio_update.txt in lib dir is missing. (please also add this for zxcvbn libs).

ximex:
I set the logging level for the php error file to `ERROR`. See: https://github.com/Admidio/admidio/pull/420/commits/f41d002c698669084f47f5dd1a9b3757289d45d1#diff-0a93838577ee904d4332f53b66bb9b3cR26
So `DEBUG` messages etc aren't logged int the system php file.

But we could use the `FingersCrossedHandler` logger. See: https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#wrappers--special-handlers

Fasse:
So were are they logged?
We should not remove the logging of our sql if I set $gDebug = 1;

ximex:
@Fasse look at: https://github.com/Admidio/admidio/pull/420/commits/f41d002c698669084f47f5dd1a9b3757289d45d1#diff-0a93838577ee904d4332f53b66bb9b3cR25

Fasse:
Log looks like this: 

> [2016-10-24 22:38:48] Admidio.INFO: SELECT *                       FROM adm_components                      WHERE  com_type = 'SYSTEM'  AND com_name_intern = 'CORE'  [] []
[2016-10-24 22:38:48] Admidio.INFO: Found rows: 1 [] []

- Why are there brackets at the end of each line?
- linebreaks are not shown. So every SQL will be in 1 line and therefore not good readable
- logfile should be in folder **adm_my_files/logs**

ximex:
@Fasse linebreaks works fine on my system

Fasse:
Maybe only on Linux?

ximex:
@Fasse Does it work if you change here the first param to this:
https://github.com/Admidio/admidio/pull/420/commits/f6af4a39dfaba683dc07750731c31507f12f4518#diff-0a93838577ee904d4332f53b66bb9b3cR27
`'[%datetime%] %channel%.%level_name%: %message% %context% %extra%\r\n'`

Fasse:
The following error occurs if you use this branch

> Notice: Undefined variable: gLogger in /Users/Jennifer/Sites/GitHub/admidio/adm_program/system/constants.php on line 104

> Fatal error: Call to a member function notice() on a non-object in /Users/Jennifer/Sites/GitHub/admidio/adm_program/system/constants.php on line 104

ximex:
@Fasse I'm a idiot ;-) in constants.php it isn't possible to log...

Fasse:
now I get the following error in the log file:
> [28-Oct-2016 06:48:17 Europe/Berlin] [2016-10-28 06:48:17] Admidio.CRITICAL: 42S22: 1054 Unknown column 'rol_administrator' in 'where clause'  {"file":"/Users/Jennifer/Sites/GitHub/admidio/adm_program/system/classes/database.php","line":705,"class":"Database","function":"showError"}

also there are no linebreaks within 1 log, so sql is not good readable. I check this with linux, mac and windows linebreaks.

ximex:
@Fasse how do you want it? you could edit the template with this string:
`"[%datetime%] %channel%.%level_name%: %message% %context% %extra%\n"`
message = The text message
context = the optional object you could add
extra = the optional object processors add (like the inspectorProcessor that adds infos where the log is executed)

ximex:
@Fasse are there any wishes open? What should i do because of the line-breaks?

Fasse:
@ximex nothing because of the line breaks, maybe I must get used with the new layout of the logging file.
But please have a look at my 2 comments above. You have removed the logging of the script calls in common.php, why? Also the line 
> ----------------------
should be added so it is vissible in the log that a new script starts here.

ximex:
@Fasse The logging of the script calls moved into `logging.php`. But they only get logged if in debug mode. Should i change this? I could also add the `---------` line.
Should i use `\n` or `\r\n` linebreaks?

Fasse:
@ximex did not saw this, that you add the lines to logging.php. Yes please add `####` line. Only for better reading.
Yes in my opinion we don't need `if (!isset($gLogger))`.
For me every log entry is in 1 line, so l don't understand what you mean with the linebreaks.

ximex:
@Fasse i mean the linebreaks between each log entry. You wrote above that in windows all logs are in one line.

Fasse:
@ximex that was a missunderstanding, the linebreaks within 1 log entry are gone. Linebreaks between each log entry are still there.

Now you can merge.

PR #432, opened 2016-10-28
Fix for getPreviousUrl

rmbinder:
In the getPreviousUrl is this code:
```
$count = count($this->urlStack);
...
// Only one url, take this one
$entry = min(0, $count - 2);
return $this->urlStack[$entry]['url'];
```

1. If only one url is in the stack, then $entry is -1 
2. If more than one url is in the stack, $entry is always 0


ximex:
i fix it directly. Thanks!
https://github.com/Admidio/admidio/commit/7b77894e9d620b012a7d9bd805e0002df7b51239

PR #436, opened 2016-10-29
improve url and path constants

ximex:
@Fasse what do you think about that?

I think we didn't need `$g_root_path` now any longer.


ximex:
@Fasse i'm finish here. feel free to merge ;-)

ximex:
@Fasse i'm finish here. Waiting for input or feel free to merge

Fasse:
@ximex I like this change. Thought about this in the past, too, but never had the time to have a look at this.

But we should not remove the $g_root_path in version 3. We should not change the config within installation and update. internal we should change everything as you do. But not in example config and installation process. Thsi should than be removed with version 4.

ximex:
@Fasse feel free to merge

PR #459, opened 2016-12-06
Feature update demo data 

Fasse:
Update demo data to version 3.2


Fasse:
@ximex can you please review

ximex:
only looked at the changes (no tests). looks good. But shouldn't have all tables have defined `auto_increment = 1`?

Fasse:
We don't need that auto_increment = 1, because we create a new table so auto_increment start value should always be set to 1. I will remove this line for 3.3

Issue #486, opened 2017-01-03
Copy user to create a new user


Fasse:
We should add the **copy** functionality to users.

Therefore add the button in user management and copy all fields except surname and firstname

PR #490, opened 2017-01-03
Copy user to create a new user 

Fasse:
#486

### DIFF
--- a/adm_program/modules/members/members_data.php
+++ b/adm_program/modules/members/members_data.php
@@ -328,20 +328,9 @@ while($row = $mglStatement->fetch())
         }
     }
 
-    if(strlen($row['email']) > 0)
-    {
-        if($gPreferences['enable_mail_module'] != 1)
-        {
-            $mailLink = 'mailto:'.$row['email'];
-        }
-        else
-        {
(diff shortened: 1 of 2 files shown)


adm_program/modules/members/members_data.php
```diff
@@ -328,20 +328,9 @@
...
-        if($gPreferences['enable_mail_module'] != 1)
-        {
-            $mailLink = 'mailto:'.$row['email'];
-        }
-        else
-        {
-            $mailLink = ADMIDIO_URL.FOLDER_MODULES.'/messages/messages_write.php?usr_id='.$row['usr_id'];
-        }
-
-        $userAdministration .= '<a class="admidio-icon-link" href="'.$mailLink.'"><img src="'.THEME_URL.'/icons/email.png"
```
ximex: do we want to remove the shortlink to send an email?

Fasse: is it useful?

I thought to much buttons are bad but if you @thomas-rcv @sirnone  also think we shouldn't remove the email link in the user management than we don't

ximex: @Fasse I do not care about this link. I only asked if this wasn't a failure

Fasse: I don't want to have too much functions and this function will be in this context not relevant.

ximex:
@Fasse you could merge

Issue #489, opened 2017-01-03
Rename user field "Address" to "Street"


Fasse:
With **Address** you normaly mean the full address of someone e.g. Street, Postcode, City.

We want the user to enter the **Street**.

PR #491, opened 2017-01-03
Rename user field "Address" to "Street" #489

Fasse:
#489


adm_program/installation/db_scripts/update_3_3.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<update>
+    <step id="10">UPDATE %PREFIX%_user_fields SET usf_name_intern = 'STREET' WHERE usf_name_intern LIKE 'ADDRESS'</step>
```
ximex: why `LIKE` an not `=`

Fasse: thought LIKE should be used with strings and = should be used with int but that was wrong.

PR #508, opened 2017-01-19
Feature prepared statements

### LINKED ISSUE #244: Prepared Statements
https://secure.php.net/manual/de/pdo.prepared-statements.php

ximex:
fixes #244 
i plan for a later version (v4.0 or later) to use/change to: https://github.com/paragonie/easydb

There are some more improvements possible. Search for comment: `// TODO add more params`


Fasse:
We should think about the placeholder. In my opinion ? will be not significant enough. We should to something like `#param#`, so its more visible.

ximex:
@Fasse the `?` is the way to do it. The php interpreter looks for a `?`. See: https://secure.php.net/manual/de/pdo.prepared-statements.php

ximex:
@Fasse fixed

Fasse:
@ximex you can merge

ximex:
@Fasse no changes of the `?` to `:name` etc?

Fasse:
In my opinion we should. But maybe we merge and changed this later?!?

ximex:
Ok. I look for this in v4.0

Issue #515, opened 2017-01-22
Enhance the layout of assigned roles rights within the profile 


Fasse:
Within the profile we Show the assigned roles rights of the User. The Layout should be enhanced and use the popover of bootstrap. Also the *assigned from* Box should be Removed.

PR #516, opened 2017-01-23
Feature profile bootstrap improvement

Fasse:
#515


ximex:
Only question is: The request for the change ask to reduce the needed space. With this solution we need more space. But it is nicer

Fasse:
Yes, we need more space for the Admin, but less space for the common user. I thought about that, too. But in my opinion you now have also more visible information than before.

PR #564, opened 2017-03-19
211 deadline for participation

thomas-rcv:
Basic implementation from wilddom´s fork:

https://github.com/wilddom/admidio/commit/a22ca0eca6076095c395bb9f3d30b61f87b95fce

Adapted to current develop branch with adjustments.


Fasse:
hmm maybe it is a problem with the event I already created. Because now this error also occurs if I open the event list.

thomas-rcv:
Changed requests. Please have a look again.

Fasse:
I have change this, so you can merge.

thomas-rcv:
This was my first Intension as well. A simple bool check. But I´m not sure we might Need the date in future.
So to encapsulate this and also the date object comparison an preventing redunat code in several scripts was a ggod idea for me.
The on top bool check is great.

PR #579, opened 2017-04-07
Improve bootstrapping

ximex:
- Separate autolading from admidio functions
- Separate non DB bootstrapping things
- Simplify bootstrapping in different files with new bootstrap file

### DIFF
--- a/adm_program/system/bootstrap.php
+++ b/adm_program/system/bootstrap.php
@@ -1,0 +1,55 @@
+<?php
+
+// embed config and constants file
+$rootPath = substr(__FILE__, 0, strpos(__FILE__, DIRECTORY_SEPARATOR . 'adm_program'));
+require_once($rootPath . '/adm_program/system/init_globals.php');
+require_once($rootPath . '/adm_program/system/constants.php');
+
+// check PHP version and show notice if version is too low
+if (version_compare(PHP_VERSION, MIN_PHP_VERSION, '<'))
+{
+    exit('<div style="color: #cc0000;">Error: Your PHP version ' . PHP_VERSION . ' does not fulfill
+        the minimum requirements for this Admidio version. You need at least PHP ' . MIN_PHP_VERSION . ' or higher.</div>');
+}
+
+// includes WITHOUT database connections
+require_once(ADMIDIO_PATH . '/adm_program/system/autoload.php');
+require_once(ADMIDIO_PATH . FOLDER_LIBS_SERVER . '/htmlawed/htmlawed.php');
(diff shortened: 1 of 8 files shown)


adm_program/system/bootstrap.php
```diff
@@ -0,0 +1,64 @@
...
+ini_set('log_errors', '1');
+
+if ($gDebug)
+{
+    ini_set('display_errors', '1');
+    ini_set('display_startup_errors', '1');
+}
+else
+{
+    ini_set('display_errors', '0');
```
Fasse: Within the else case we should do nothing and use the default values of the system. If someone want's to view the errors (maybe in a test area) so why should we set this to 0?

adm_program/system/function.php
```diff
@@ -9,51 +9,9 @@
...
-    }
-
-    return false;
-}
-
-// now register this function in this script so only function.php must be included for autoload
-spl_autoload_register('admFuncAutoload');
+// Load autoloading file
+// Deprecated: Backwards compatibility
+require_once(ADMIDIO_PATH . '/adm_program/system/autoload.php');
```
Fasse: I think we can remove this because within plugins or so you include the common.php and there is now the new autoload included.

ximex: i don't know if someone doesn't use the `common.php`. But if you think we can remove it, i will do it.

adm_program/system/bootstrap.php
ximex: Look at: http://www.phptherightway.com/#error_reporting
In a production system you shouldn't show error messages to the user. This are important infos for a hacker. If you have a test system you should always set `$gDebug = true`. So we shouldn't change here anything

Fasse: you are right, but you don't know if the user use our system in production or test enviroment. So it's up to the user/server admin to set these variables. If I have a test server with x scripts I don't want to set in each script a "debug" mode. I set the server parameters as I think they are right.

ximex: Ok. Idea:
`$gDebug` is not set -> use default
`$gDebug` is true -> set values to `'1'`
`$gDebug` is false -> set values to `'0'`

adm_program/system/function.php
ximex: So what should i do?

Fasse: yes

adm_program/system/bootstrap.php
Fasse: I thought also about this, we can also leave the settings as they are because we now have a good logging.

Issue #587, opened 2017-04-17
Wrong database object set within update to version 2.4


Fasse:
view forum https://www.admidio.org/forum/viewtopic.php?f=3&t=7499

PR #589, opened 2017-04-18
update fix

wobsta:
https://www.admidio.org/forum/viewtopic.php?f=3&t=7499#p26250


Fasse:
Thanks for the fix, but I already fix it with https://github.com/Admidio/admidio/issues/587 .

wobsta:
sorry for the noise, I didn't notice as I just checked the master branch, not v3.2 ...

PR #597, opened 2017-04-23
573 possibility for a leader to edit status

thomas-rcv:
- Outsourced modal window in own file
- New link for leaders and Admins to edit participation states of users
- some more logic and logic changes


thomas-rcv:
Problem fixed and moved file to module dates

adm_program/modules/lists/lists_show.php
```diff
@@ -389,6 +402,7 @@
 
         // create html page object
         $page = new HtmlPage();
+        $page->enableModal();
```
Fasse: Only enable if $editUserStatus = true

thomas-rcv:
You´re right. Modal is now only allowed with permission to edit.
Correct message should also be shown now without any changes, except for the comments.
This, I want to keep at this time to enable documentation by leaders and admins afterwards, if deadline exceeded.

Notice.:
Deadline exceeded is also working within the PR. My data build is as old as it can show it up directly.
I think no tests are needed for this.

Fasse:
Everything works fine :)

You can merge.

Issue #199, opened 2015-10-23
Assign roles to categories to manage the visiblity


Fasse:
We should allow to assign multiple roles to each category. These roles have the permission to view this category and all of their items. If no roles are assigned than all users can see that category.

With this we can manage to show profile fields only for some roles. 
This feature could also be interesting in other modules with categories.
Within the event module we can handle the existing **visible for** with this new feature and use the visible for as a assignment for roles that are allowed to participate to an event.

PR #614, opened 2017-05-18
Feature categories roles rights

Fasse:
#199


thomas-rcv:
Did first test without knowledge of the code:

- First,I was confused where to find the new category settings. It took some time to find in the individual module settings. Maybe we should think about the wording in the role categories where I expected to find.
Within the new feature the name "Role Groups" would be more especailly and independant from new categories.

- Within Events, I get the message that the users has no rights to view the category when I activte the participation option on creation/editing..I expected to have the option "all (all visitors)" if the calendar is configured like that. Choosing anything it runs in the message.

Fasse:
- You are right the category settings are very much hidden. But this was not changed within this feature request. It's still the same place like in 3.2 . But maybe we should change it like it's for the "Vorstand". Please login as "Vorsitzender" and have a look at it.
- I fix the error within the event module.

Fasse:
I have add the link to manage categories to every module menu. Now it should be more intuitive to get to the category configuration.

thomas-rcv:
The new links helps. It is ok
The message it is also solved.

Did some further test:
The Administrator sees everything and is always able to participate himself. I thought the role Administrator follows also the rules of visibilty. I think this was the behaviour in the past.
Maybe the Administrator is not a member of the organization and only manages configuration stuff?
In my point of view he should not be able to see the calender for explicit defined roles in which he doesn´t participate ( e.g. Managing Board ).

Editing/Creating an Event I get the message, that the field participants is empty if I doesn´t select. Would it not be an idea to allow this also and set all visible roles defined for the calendar to possible to participate?

If I select a role for participation I get and empty select2-change link in the field (<li class="select2-selection__choice" title=" ">). This should be checked.

A helptext for the new particpation input field for roles with an explantation of this field would be great.

The calendar configuartion works like expected for role settings

ximex:
@Fasse please use `replaceValuesArrWithQM` for `SQL IN (...)` queries

adm_program/modules/categories/categories_function.php
```diff
@@ -173,6 +181,22 @@
...
     }
 
+    // roles have their own preferences for visibility, so only allow this for other types
+    // until now we do not support visibility for categories that belong to several organizations
+    if($getType !== 'ROL'
+    && ($category->getValue('cat_org_id') > 0 || ((int) $category->getValue('cat_org_id') === 0 && $gCurrentOrganization->countAllRecords() === 1)))
+    {
+        // save changed roles rights of the category
+        $rightCategoryView = new RolesRights($gDb, 'category_view', $category->getValue('cat_id'));
+        $rightCategoryView->saveRoles($_POST['adm_categories_view_right']);
```
thomas-rcv: $_POST['adm_categories_view_right'] must be validated!

adm_program/modules/dates/dates_function.php
```diff
@@ -89,9 +89,21 @@
...
     // ------------------------------------------------
 
+    if(!isset($_POST['date_registration_possible']))
+    {
+        $_POST['date_registration_possible'] = 0;
+    }
+    if($_POST['date_registration_possible'] == 1
+    && (!isset($_POST['adm_event_participation_right']) || array_count_values($_POST['adm_event_participation_right']) == 0))
+    {
+        $_SESSION['dates_request']['adm_event_participation_right'] = '';
```
Fasse: Until now we have no function to validate arrays. Had discuessed this with ximex before. We want to integrate a lib for validation after that we could also validate arrays. But since my last commit all methods in RolesRights will check the array for integer values. So everything with class RolesRights is save.

adm_program/modules/categories/categories_function.php
```diff
@@ -109,20 +109,26 @@
         // => EXIT
     }
 
+    if($getType !== 'ROL' && ((bool) $category->getValue('cat_system') === false || $gCurrentOrganization->countAllRecords() === 1) && !isset($_POST['adm_categories_view_right']))
```
Fasse: done

adm_program/modules/categories/categories_function.php
```diff
@@ -173,6 +181,22 @@
         // => EXIT
     }
 
+    // roles have their own preferences for visibility, so only allow this for other types
+    // until now we do not support visibility for categories that belong to several organizations
+    if($getType !== 'ROL'
+    && ($category->getValue('cat_org_id') > 0 || ((int) $category->getValue('cat_org_id') === 0 && $gCurrentOrganization->countAllRecords() === 1)))
```
Fasse: done

adm_program/modules/dates/dates_function.php
```diff
@@ -293,6 +294,21 @@
...
 
+    if(isset($_POST['adm_event_participation_right']))
+    {
+        // save changed roles rights of the category
+        $rightCategoryView = new RolesRights($gDb, 'category_view', (int) $_POST['dat_cat_id']);
+
+        // if roles for visibility are assigned to the category than check if the assigned roles of event particiaption
+        // are within the visibility roles set otherwise show error
+        if(count($rightCategoryView->getRolesIds()) > 0
+        && count(array_intersect($_POST['adm_event_participation_right'], $rightCategoryView->getRolesIds())) !== count($_POST['adm_event_participation_right']))
```
Fasse: here is 1 condition per line, nothing to change

Fasse:
@ximex thanks for the hint. I will use this function in future developement.

Fasse:
@thomas-rcv 

> The Administrator sees everything and is always able to participate himself. I thought the role Administrator follows also the rules of visibilty. I think this was the behaviour in the past.
> Maybe the Administrator is not a member of the organization and only manages configuration stuff?
> In my point of view he should not be able to see the calender for explicit defined roles in which he doesn´t participate ( e.g. Managing Board ).

The visibility is now controlled by the categories. If you select visibility roles for the categories the administrator doens't belong than he couldn't view any events of that category.
The same with the participation. If he doesn't belong to any role than he could not participate.

> Editing/Creating an Event I get the message, that the field participants is empty if I doesn´t select. Would it not be an idea to allow this also and set all visible roles defined for the calendar to possible to participate?

I doesn't like this idea because than we have a special meaning for no role selected. Now the configuration is very straight with no exceptions. If we implment this than we must implement an exception at every part where we use RolesRights.

> If I select a role for participation I get and empty select2-selection_choice link in the field (<li class="select2-selection__choice" title=" ">). This should be checked.

Could you give me some more informations how to reproduce this. In my first tests I could not reproduce this.

> A helptext for the new particpation input field for roles with an explantation of this field would be great.

Do you really think this is possible. I thought **Participiants** and a roles selectbox will explain itself. I don't want every field with a explaination. But maybe I'm already too much in the process and an explaination is useful.

I also check all your code comments and fix them if possible.

thomas-rcv:
> The visibility is now controlled by the categories. If you select visibility roles for the categories the administrator doens't belong than he couldn't view any events of that category.
 The same with the participation. If he doesn't belong to any role than he could not participate.
 Or maybe I don't understand your problem?!

My mistake - Forget this issue
Paul Webmaster has also role Vorsitzender. This is the reason why my test failed

> I doesn't like this idea because than we have a special meaning for no role selected. Now the configuration is very straight with no exceptions. If we implment this than we must implement an exception at every part where we use RolesRights.

Is ok for me. We might define this field as mandatory field

> Could you give me some more informations how to reproduce this. In my first tests I could not reproduce this.

Try to force the message  " Feld Teilnehmer ist nicht gefuellt" After jumping back to edit the event choose a role. After that you will habe the empty enty with only a delete Icon.
Maybe this could be solved if defining the field as mandatory ...

> Do you really think this is possible. I thought Participiants and a roles selectbox will explain itself. I don't want every field with a explaination. But maybe I'm already too much in the process and an explanation is useful.

With Label "Teilnehmer" and role selectiobox I understand that These roles will participate ( are assigned). So one way, a clear explantion with a helptext, or maybe much better take new text string "Anmeldung möglich für:" So, it would also be quiet clear instead using helptext.


The code updates and especially the callback function are ok for me and I will approve the review now.
After you got the empty choice_selection entry you could merge.
Also I would appreciate thinking about the label/helptext for the participants field. It is not an error, just an point of view to improve.

ximex:
@Fasse please wait with the merge. i will do a second review of this code

Fasse:
@thomas-rcv I have renamed the label. 
I think I found the problem with the empty entry. Can you please retest this?

thomas-rcv:
@Fasse 

Looks good now.
You can merge.

Fasse:
@ximex can I merge?

ximex:
@Fasse yes

PR #656, opened 2017-10-28
add string helper functions

ximex:
so the checks would be simpler and easier to read what they do


ximex:
Maybe change to function name to `admIsTranslationStrId`?

Fasse:
Yes `admIsTranslationStrId` would be better.

I don't have checked all replacements but I checked your new functions and I like them. The code will be more readable.

So you can merge after rename of the function.

Issue #514, opened 2017-01-22
Update Cookies after Admidio Update


ximex:
https://www.admidio.org/forum/viewtopic.php?f=7&t=7323

PR #661, opened 2017-11-01
Feature make database serializeable

### LINKED ISSUE #660: Make Database class serlalizeable
Make the database class serializeable with reconnection so we don't have to reset the database object in other classes

ximex:
fixes #660

### DIFF
--- a/adm_program/system/classes/database.php
+++ b/adm_program/system/classes/database.php
@@ -95,10 +95,6 @@ class Database
      */
     protected $dsn;
     /**
-     * @var string
-     */
-    protected $dbEngine;
-    /**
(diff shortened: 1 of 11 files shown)


ximex:
one problem what could happen: you have to logout and relogin to renew the session objects that are incompatible.
I think this is the same problem: https://github.com/Admidio/admidio/issues/514

adm_program/system/classes/database.php
```diff
@@ -64,6 +64,10 @@ class Database
     /**
      * @var string
      */
+    protected $engine;
```
Fasse: Please add one line description with example value.

PR #662, opened 2017-11-01
Feature settings manager

### LINKED ISSUE #659: Refactor $gPreferences in a SettingsManger class

ximex:
fixes #659


ximex:
One step is missing:
Use of the methods `getBool()`, `getInt()`, `getFloat()`, `getString()` to get values in correct types

ximex:
The last 143 `$gSettingsManager->get('xxx_xxx_xxx')` are from int values that are hard to see what the values mean. I think we should change this settings to strings that are readable.
But i could also change them to int for now

ximex:
Please wait with merge. I will change some parts to implement the PSR-16 interface.
http://www.php-fig.org/psr/psr-16/

ximex:
Ok it doesn't work as expected. Forget the last message.

ximex:
@Fasse any questions to this commit? I'm now finished

Fasse:
I could not call any page on this branch. Everything is broken :(

ximex:
Ok this is a session problem. If you delete the session cookie it should work. This should be the same problem: https://github.com/Admidio/admidio/issues/514
I will look to add a session clean or something else on updates

Fasse:
hmm, I'm not able to perform an update from 3.2 to 3.3. It's still the same problem as before:

> Fatal error: Uncaught UnexpectedValueException: Settings name "enable_gbook_comments4all" is an invalid string! in /admidio/adm_program/system/classes/settingsmanager.php:318 Stack trace: #0 /adm_program/installation/update.php(360): SettingsManager->setMulti(Array, false) #1 {main} thrown in /admidio/adm_program/system/classes/settingsmanager.php on line 318

ximex:
The only problem i got on the update is the problem with the cookies/sessions. I deleted the cookies and than refreshed the page and could start the update.
Look here: https://github.com/Admidio/admidio/issues/514

ximex:
@Fasse please test. Maybe the update problem is now also gone

Issue #583, opened 2017-04-13
Dynamic count of translation params


ximex:
Now it is fixed to a max of 4 params.

Fasse:
Why do you don't create a pull request?

ximex:
My fault. Thought that this changes are small enouth that no PR is required

ximex:
Do we want to allow if you have only one additional translation param that you could add it as string?
Examples:
```php
$gL10n->get('XXX_XXX_XXX', array('param1')); // already possible
$gL10n->get('XXX_XXX_XXX', array('param1', 'param2')); // already possible
$gL10n->get('XXX_XXX_XXX', 'param1', 'param2'); // already deprecated
$gL10n->get('XXX_XXX_XXX', 'param1'); // allow again <<<------
```

PR #667, opened 2017-11-03
Feature improved dynamic language params

ximex:


ximex:
see #583

sirnone:
I don't like the possibility to use array or String - this makes this easy call complicated without an need for it!?

Just array would be better also vor readability i think.

Fasse:
My opinion:
For compatibility we should allow string in 3.3 as well as array. But array should the favorit and in 4.0 we remove the string option.

Admidio core should also in 3.3 only use arrays.

ximex:
Ok than i will change it that admidio uses only arrays intern but also support strings for backwards compatibility

ximex:
@Fasse feel free to merge

PR #670, opened 2017-11-07
Feature disk space

ximex:
Adds a progress-bar of the used disk space

### DIFF
--- a/adm_program/system/function.php
+++ b/adm_program/system/function.php
@@ -697,9 +697,9 @@ function admFuncGetDirectoryEntries($directory, $searchType = 'file')
 }
 
 /**
- * Gets the total, free and used disk space.
+ * Gets the total, free and used disk space in bytes.
  * @param string $path Path of the filesystem
- * @return array<string,int>
+ * @return array<string,int> Return format: array("total" => $total, "free" => $free, "used" => $used)
  */
 function admGetDiskSpace($path = '/')
 {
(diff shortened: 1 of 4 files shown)


adm_program/system/function.php
```diff
@@ -697,6 +697,48 @@ function admFuncGetDirectoryEntries($directory, $searchType = 'file')
 }
 
 /**
+ * Gets the total, free and used disk space.
+ * @param string $path Path of the filesystem
+ * @return array<string,int>
```
Fasse: Please add a description here how the array will be filled with each value. Maybe additional an example.

adm_program/system/function.php
```diff
@@ -697,6 +697,48 @@ function admFuncGetDirectoryEntries($directory, $searchType = 'file')
...
+    $used = $total - $free;
+
+    return array('total' => $total, 'free' => $free, 'used' => $used);
+}
+
+/**
+ * Returns a human readable size with unit
+ * @param int  $size The size in bytes
+ * @param bool $si   Use SI or binary unit. Set true for SI units
+ * @return string
```
Fasse: Please add a description here how the array will be filled with each value. Maybe additional an example.

adm_program/system/function.php
Fasse: It's a short function and you may see it in the code but you should also have in mind that we can create a documentation without the code and than you don't know how the array looks like.

adm_program/system/function.php
ximex: which array? I think you mean the format of the return string

Fasse: ok, copy paste thing. I mean the return string.

ximex:
@Fasse there is only a problem if we have a so long text with the min-width. Look here: https://getbootstrap.com/docs/3.3/components/#progress-label

Fasse:
Ok, I understand. But maybe we should rename it to "Speicherplatz" anyway?

PR #672, opened 2017-11-13
Feature safeurl

### LINKED ISSUE #664: Add function to build safe urls
add a function to build safe urls where all params are urlencoded and the full string is encoded to put it save in html code

ximex:
fixes #664


ximex:
I'm finished here. Please review

ximex:
@Fasse please test now. I think i found where the problem is.
If you we don't found url errors now i will switch the escape flag back to true and look at the special cases with the errors

Fasse:
Still the same error. Did you have done an update successful?

ximex:
@Fasse found the stupid bugs. Please test and merge if all is ok

PR #704, opened 2018-01-17
Feature use filesystemutils

ximex:
use the new FileSystemUtils class and some code improvements


ximex:
@Fasse please review and merge if all is ok

Fasse:
@ximex before I start testing this, have you manually test your changes. Because this is something that could break our file handling ...

Fasse:
I will check this after commit of menu so I could check compatibility also against those changes.

ximex:
@Fasse now you could test. Fixed some small bugs

Fasse:
There is a problem with the htaccess file. If the file doesn't exists, it won't be created with the new class. So the folder isn't protected.

ximex:
On my server the creation of `.htaccess` works

Fasse:
I could reproduce it with my Mac and my Linux NAS.

To reproduce:
I have a running Admidio installation folder adm_my_files and a htaccess file.
Then I remove the htaccess file.
Now go to downloads and create new folder 
-> no htaccess file is created

ximex:
only on update now. Look here: https://github.com/Admidio/admidio/blob/feature-use-filesystemutils/adm_program/installation/update_functions.php#L237
I could also add this code into common.php than it would be checked on any request. but i think on update it should be enough.

Fasse:
Hmm, of you move your Installation to another webspace it should also be added. Check this at every request would be too mich, but maybe if we still handle with adm_my_files folder.

ximex:
Or show it in the system-settings and add a button to create on if there is no `.htaccess`

Fasse:
Ok, also possible. But Check at update should still be there.

ximex:
Ok than i will implement this

ximex:
@Fasse done

adm_program/installation/update.php
```diff
@@ -102,11 +102,27 @@
...
 
+if (!is_executable(ADMIDIO_PATH . FOLDER_DATA) || !is_writable(ADMIDIO_PATH . FOLDER_DATA))
+{
+    try
+    {
+        FileSystemUtils::chmodDirectory(ADMIDIO_PATH . FOLDER_DATA, 0777);
+    }
+    catch (\RuntimeException $exception)
+    {
+        // TODO
```
Fasse: We should leave a message that the user must set the rights to **adm_my_files**.

adm_program/languages/en.xml
```diff
@@ -1135,6 +1135,8 @@
   <string name="SYS_DELETED_USER">Deleted user</string>
   <string name="SYS_DESCRIPTION">Description</string>
   <string name="SYS_DESIGNATION">Name</string>
+  <string name="SYS_DIRECTORY_PROTECTION">Directory protection</string>
```
Fasse: Just for info: you must only change the de.xml . Thats the only one that will be uploaded to transifex. All other files will than be downloaded from transifex.

ximex:
@Fasse couldn't reproduce your problem. Please give more information.

Fasse:
you can merge.

Issue #606, opened 2017-05-03
System notifications got 2 new lines if 1 line was defined


Fasse:
If you define a system notification within the preference like this

> Test 1
> Test 2
> Test 3

Then you got the mail like this

> Test 1
>
> Test 2
>
> Test 3
>

PR #708, opened 2018-01-21
fixed duplicate line breaks in email messages

deki:
similar to the fix in #606

### DIFF
--- a/adm_program/system/classes/email.php
+++ b/adm_program/system/classes/email.php
@@ -279,12 +279,8 @@ class Email extends PHPMailer
     public function setText($text)
     {
         // Erst mal die Zeilenumbrueche innerhalb des Mailtextes umwandeln in einfache Umbrueche
-        // statt \r und \r\n nur noch \n
-        $replaceArray = array(
-            "\r"   => "\n",
-            "\r\n" => "\n"
-        );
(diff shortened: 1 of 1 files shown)


adm_program/system/classes/email.php
```diff
@@ -280,11 +280,7 @@ public function setText($text)
     {
         // Erst mal die Zeilenumbrueche innerhalb des Mailtextes umwandeln in einfache Umbrueche
         // statt \r und \r\n nur noch \n
-        $replaceArray = array(
-            "\r"   => "\n",
```
ximex: @deki could you test it if it also works if you change line 284 and 285? I think this only is the wrong order

deki: You are right but shouldn't we use the same fix for all occurences? That's why I adapted the fix that was applied to systemmail.php in #606.

Fasse: I also don't understand why we must replace `\r`. 
I know `\n` for Linux and `\r\n` for Windows.

ximex: hmmm... right. why have we done this?

ximex:
@deki could you also change the comment?

deki:
done

Fasse:
Thanks for the fix

PR #716, opened 2018-02-07
add and use new TABLE_PREFIX const

ximex:
So we need no global import any more and the table prefix should be a constant


ximex:
@Fasse feel free to merge.

ximex:
Should i also add:
- `$gDbType` => `DB_ENGINE` or `DB_TYPE`
- `$g_adm_src` => `DB_HOST` or `DB_DOMAIN`
- `$g_adm_port` => `DB_PORT`
- `$g_adm_db` => `DB_DATABASE` or `DB_NAME`
- `$g_adm_usr` => `DB_USERNAME`
- `$g_adm_pw` => `DB_PASSWORD`

Fasse:
forget my last comment ...

Fasse:
Hmm, we should not change the config.php.

PR #820, opened 2018-08-10
refactor navigation class

ximex:
Is it really necessary to clear the url stack if you go to another module? (`$nav->addFirst(...)`)


Fasse:
It was the easiest way to intialize the stack again. But if you have time to test you can change the handling so that we don't initialize the stack with every module.

ximex:
not finished yet

Fasse:
What is about this change request? Do you want to finish this?

Fasse:
@ximex What's about this pull request?

PR #855, opened 2018-12-24
Streaming chunks of files for large file downloads

gabipetrovay:
This will start large file downloads immediately after the user clicks. The `readfile` will read the entire file to download in memory first.


gabipetrovay:
This was inspired from: https://zinoui.com/blog/download-large-files-with-php

I had problem downloading 500 MB files (up to 2 minute delays before the download was actually starting). The solution proposed here fixed the issue.

A review is for sure more than welcome!

Fasse:
@gabipetrovay Thanks for the fix.

PR #880, opened 2019-04-11
Customize visibility of categories in own profile.

Laqoore:
Categories that are hidden are not displayed even if you edit your own profile.

### DIFF
--- a/adm_program/system/classes/ProfileFields.php
+++ b/adm_program/system/classes/ProfileFields.php
@@ -462,8 +462,7 @@ class ProfileFields
         // check if the current user could view the category of the profile field
         // if it's the own profile than we check if user could edit his profile and if so he could view all fields
         // check if the profile field is only visible for users that could edit this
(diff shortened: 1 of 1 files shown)


Laqoore:
Dependency to language files for 'CAT_PROFILE_FIELDS_VISIBILITY' exists

Fasse:
If we accept this patch we change the behavior of the hidden categories.

Now you can always view hidden categories in your own profile. That should be used for e.g. bank account.

If we use your patch than hidden categories can only be edited and viewed by other people. In which case do you need that?

Laqoore:
I need this e.g. for the field "deceased on" (see also https://www.admidio.org/forum/viewtopic.php?t=5258 )

The behavior does not change after my tests if the visibility of the category is set for a role of the user. This patch also allows me to completely recreate the original behavior - if desired.

Laqoore:
Okay, I see what you mean. 
With the patch it is no longer possible to set the category so that for example only administrators AND only the user himself can change the fields.

Better would be a system role "myself" or similar.

Fasse:
As describe above we could not accept this pull request because the behavior is changed.

PR #943, opened 2020-10-21
Task 942 rework password utils

ximex:


Fasse:
We used md5 in the past. If someone has his password hashed with md5, it should be possible that we could read it. I saw that you removed md5, so it isn't possible to login if you have a password with md5 hash. Is that right?

Fasse:
@ximex  I have added md5 support again. If it's ok for you I would merge it.

Issue #937, opened 2020-10-14
Replace master/slave, black/white


ximex:
Replace problematic words
https://www.heise.de/news/Nichtrassistische-Sprache-Abschied-von-Blacklist-und-Whitelist-4784291.html

Fasse:
Some of the things I found:

master (Branch)  --> should be fixed if GitHub has better migration tools, beginning 2021 --> synonym **main**
blacklist --> StringUtils class --> synonym ???
whitespace --> **blank space**, **empty space**
webmaster --> **administrator**
master data --> **main data**

ximex:
Most other projects change Blacklist to Denylist or Blocklist (Whitelist to Allowlist)

PR #984, opened 2021-01-16
Change code language blm

Fasse:
#937


adm_program/languages/de.xml
```diff
@@ -741,6 +741,7 @@
   <string name="SYS_BACKUP_FILE">Backupdatei</string>
   <string name="SYS_BACKUP_FROM">Backup vom #VAR1# um #VAR2#</string>
   <string name="SYS_BACKUP_ONLY_MYSQL">Das Backupmodul kann nur bei einer MySQL-Datenbank genutzt werden.</string>
+  <string name="SYS_BASIC_DATA">Basic data</string>
```
ximex, line 744: should be german

ximex (changes requested):
@Fasse i'm not sure about `whitespace` and `empty space`. I never heard about `empty space`. And in this case the "white" from `whitespace` depends on the white paper and the missing black ink. But not sure

adm_program/languages/de.xml
Fasse: What do you think is a good name **Basisdaten** or **Grundlegende Daten**?

ximex: `Basisdaten`

Fasse: > @Fasse i'm not sure about `whitespace` and `empty space`. I never heard about `empty space`. And in this case the "white" from `whitespace` depends on the white paper and the missing black ink. But not sure

Ok, if you like we could stay with whitespace. Give me a short feedback.

Issue #843, opened 2018-10-15
Send Testmail


ximex:
Add a button to send a testmail
https://www.admidio.org/forum/viewtopic.php?f=2&t=7854

PR #985, opened 2021-01-17
  Send Testmail #843

Fasse:
Send Testmail #843


ximex:
didn't do any real test. but review look good

PR #998, opened 2021-01-25
Dockerfile improvements

schast:
* Add docker usage help to README.md
* create README-Docker.md with detailed docker usage help

Dockerfile improvments:
* Add healthcheck
* Install postfix to support mail delivery
* Install trivy to scan for security issues during image build
* Add build lables
* Add many configuration variables
* Add custom dockerhub build hook

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,3 +10,4 @@
 /demo_data/ export-ignore
 /hooks/ export-ignore
 /dockerscripts/ export-ignore
+/Dockerfile export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,5 @@
 # exclude some folders in zip
 /.* export-ignore
 /demo_data/ export-ignore
+/hooks/ export-ignore
+/dockerscripts/ export-ignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,75 +1,6 @@
 # local image build command:
 #   IMAGE_NAME="yourUsername/admidio:v4.0.4" ./hooks/build
 
-# Docker Container Example:
-# docker run --name Admidio \
-#   --memory="1024m" --cpu-quota="80000" --cpuset-cpus="0-1" \
-#   -p 8095:8080 \
-#   --restart=unless-stopped \
-#   -v Admidio-files:/opt/app-root/src/adm_my_files \
-#   -v Admidio-themes:/opt/app-root/src/adm_themes \
-#   -v Admidio-plugins:/opt/app-root/src/adm_plugins \
-#   --link Admidio-MariaDB:mysql \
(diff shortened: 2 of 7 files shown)


ximex:
The header image in the README.md isn't visible as it is white

Fasse:
Maybe we should not make a subtitle to installation and update. We should add a new section for docker maybe after ##update and add there the docker content. 

You add some subfolders for docker `dockerscript` and `hooks` . Are they really necessary? Must we deliver these folders or could we add the files into a subfolder of the existing folders?

Fasse:
btw there is also a `admidio_logo_black_200.png` and `admidio_logo_darkblue_200.png` in the images folder.

schast:
> Maybe we should not make a subtitle to installation and update. We should add a new section for docker maybe after ##update and add there the docker content.
> 
> You add some subfolders for docker `dockerscript` and `hooks` . Are they really necessary? Must we deliver these folders or could we add the files into a subfolder of the existing folders?

The `dockerscript` folder can be moved anywhere. Any suggestions?

The `hooks` folder can not be moved, since it must be relative to the Dockerfile and overwrite the dockerhub basic build command. The `hooks` folder does not need to be delivered.

Fasse:
Ok, than we should add the `hooks` to `.gitattributes` so we don't deliver that folder like `demo_data`.

The `dockerscript` would be perfect in `adm_program/system` but if you think it's to deep and hidden, we can add it just to `adm_program`.

It would be nice if you add one or two sentence at the beginning of the docker section in the readme. Something like 

> Admidio could also be used within a docker enviroment. Therefore we deliver a Dockerfile and you can easily setup your own docker image with the following command:

Also we could add a hint to https://hub.docker.com/r/admidio/admidio/ where they can find docker images.

schast:
I don't think the right place for `dockerscripts` is in `adm_program/system` or in `adm_program` since it should only contains the admidio application code.
But the `dockerscripts` are only for docker administration.
Would it be sufficient to include `dockerscripts` also in .gitattributes?

Fasse:
> Would it be sufficient to include `dockerscripts` also in .gitattributes?
If so than we wouldn't deliver these script with our download package and we could only link to those scripts in our doku or readme. If it's ok for you than we could do it. 

I don't want to deliver a new folder because users then have another folder that they must have to look for and update this folder ...

schast:
I think in the ZIP delivery the Docker components do not need to be provided.

Maybe it would be nice to add the docker help also in the admidio wiki (https://www.admidio.org/dokuwiki/doku.php?id=de:2.0:installation)?

Fasse:
Yes, it would be nice to have a docker page in our wiki. 

You can register at the wiki and send me your username. I will than give you the rights.

PR #1002, opened 2021-01-28
Docker-compose example should be list

c-holtermann:


Fasse:

PR #1003, opened 2021-01-28
add option to disable ipv6

c-holtermann:
admidio container kept restarting for me in my docker-compose setting. I found that was because postfix did not start. It showed `postfix:` fatal: parameter inet_interfaces: no local interface found for ::1`. That error message was only shown when I started the container with `docker-compose -p PROJECTNAME run admidio`. It would be good to easily document how to have errors shown and how to debug (but maybe I just didn't look good enough).

In my setting obviously ipv6 is disabled. I don't use it.

To keep things simple for me I added an environment option ADMIDIO_NO_IPV6:

    environment:
       ADMIDIO_NO_IPV6: "YES"

If this is true:
* `inet_interfaces = all` will be commented out
* `inet_protocols` will be set to `ipv4`

It may be helpful to check that in runtime, in my case checking this would have worked in the container:

    # sysctl -a 2>/dev/null | grep disable_ipv6
    net.ipv6.conf.all.disable_ipv6 = 0
    net.ipv6.conf.default.disable_ipv6 = 0
    net.ipv6.conf.eth0.disable_ipv6 = 0
    net.ipv6.conf.eth1.disable_ipv6 = 0
    net.ipv6.conf.lo.disable_ipv6 = 0

From: https://www.golinuxcloud.com/linux-check-ipv6-enabled/

I did not check how this environment option fits with the others and how checking for true/false is adequately done in the current setting.

### DIFF
--- a/dockerscripts/startup.sh
+++ b/dockerscripts/startup.sh
@@ -16,9 +16,16 @@ set -o pipefail  # causes a pipeline (for example, curl -s https://sipb.mit.edu/
 # configure apache
 sed -i "s/#ServerName.*/ServerName \$\{HOSTNAME\}/g" /etc/httpd/conf/httpd.conf
 
+if [ "${ADMIDIO_NO_IPV6}" != "" ]; then
+  sed -i "s/inet_interfaces = all/#inet_interfaces = all/g" /etc/postfix/main.cf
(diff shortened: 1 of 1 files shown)


schast:
should be fixed with pr #1010

Fasse:

PR #1010, opened 2021-01-31
Detect ipv6 status in startup.sh for Docker

schast:
Detect ipv6 status and set inet_protocols to ipv4 if disabled to prevent "no local interface found" error on container startup; Add some logging output for better debugging (#1003)


Fasse:

PR #1041, opened 2021-05-09
CSV Import: Try to auto-detect the correct columns (compare CSV column headers with internal and visible field names)

kainhofer:


Fasse:
I have changed the layout of the message which columns of the file were not assigned. Maybe we should not show that box if all columns are assigned.

kainhofer:
> I have changed the layout of the message which columns of the file were not assigned. Maybe we should not show that box if all columns are assigned.

That message looks fine. I thought about hiding the box, too, but then decided not to. My rationale was that when hiding the box, the form contents under the mouse unexpectedly shift up/down when the user works on the form, which is bad usability in my eyes.
Of course, if the info box would be an overlay and not part of the normal page flow, then showing/hiding would not have an influence on the form display/position, then things would be different. But I can't think of a good way to turn the box into an overlay.

If you un-comment the show/hide calls in the JS in my last commit, you can see the form behaviour when hiding the box.

Fasse:
Ok, that looks fine so. 

One minor note: could you give id and class names the prefix _admidio_
e.g. _admidio-import-unused_

We use this in newer changes so you know if it's an admidio class or something from bootstrap. Also with the id you know if it's an admidio element.

After that change we can merge this feature request.

adm_program/languages/en.xml
```diff
@@ -1403,4 +1404,5 @@
   <string name="SYS_YOUR_LISTS">Own lists</string>
   <string name="SYS_YOUR_NAME">Your name</string>
   <string name="SYS_Z_TO_A">Z to A</string>
+  <string name="MEM_POSSIBLE_FIELDNAMES">Column names for autodetection: #VAR1#, #VAR2#</string>
```
Fasse, line 1407: could you change the position in file so this string will be by the other MEM strings.

adm_program/modules/members/import_csv_config.php
```diff
@@ -86,8 +86,36 @@
 // show form
 $form = new HtmlForm('import_assign_fields_form', ADMIDIO_URL. FOLDER_MODULES.'/members/import_csv.php', $page, array('type' => 'vertical'));
 $form->addCheckbox('first_row', $gL10n->get('MEM_FIRST_LINE_COLUMN_NAME'), $formValues['first_row']);
+$form->addHtml('<div class="alert alert-warning alert-small" id="import-unused"><i class="fas fa-exclamation-triangle"></i>'.$gL10n->get('MEM_IMPORT_UNUSED_HEAD').'<div id="import-unused-fields">-</div></div>');
```
Fasse, line 89: please add admidio prefix to id

adm_program/modules/members/import_csv_config.php
```diff
@@ -86,8 +86,36 @@
...
+        }
+        $("#import-unused #import-unused-fields").html(outstr);
+    });
+    $(".import-field").trigger("change");',
+    true
+);
+
 $htmlFieldTable = '
-    <table class="table table-condensed">
+    <table class="table table-condensed import-config import-config-csv">
```
Fasse, line 118: please add admidio prefix

kainhofer:
I have now added "admidio-" prefixes to the CSS classes and IDs. And I moved the MEM_ string to its proper position in the language files.

Fasse:
Thanks for implementing this feature!

Issue #826, opened 2018-09-12
Improve Import


ximex:
- Use lib to parse Files
- Options for delimiter, enclosure, encoding
- Options for match columns (https://www.admidio.org/forum/viewtopic.php?f=2&t=7621)
- Add support for .ods, .xlsx (With new lib)

PR #1043, opened 2021-05-11
Import: Use the phpSpreadsheet library to import data from xls(x), ods, csv, html, etc.

kainhofer:
Import: Use the phpSpreadsheet library to import data from xls(x), ods, csv, html, etc.

-) Allow choosing the import file type (auto-detect, xlsx, xls, ods, csv, html (and all other formats auto-detected by phpSpreadsheet))
-) Depending on the import file type, provide settings for the worksheet inside the spreadsheet, or the encoding, separator and enclosure of CSV-files (use javascript to show/hide and en/disable the unused settings for a format)
-) Use the phpSpreadsheet library to load all types of import file, using the settings selected by the user. The data will be available as an array, irrespective of the original import format
-) The actual import function is unchanged and simply uses the array now created by phpSpreadsheet (previously by parsing the CSV lines manually)

This implements parts 1, 2 and 4 of Issue #826


Fasse:
That looks like a good feature. I will review this at the weekend.

One thing that I saw in the files is about the translation string. You use for example `<string name="SYS_CSVIMPORT_SEP_AUTO">Auto-detect</string>` . We try to use the full english name within the name attribute so you better know in the code which text should be shown. Also the context information CSVIMPORT is not necessary because we want to use the same text also in other modules if it's useful there. So a better name for it would be `<string name="SYS_AUTO_DETECT">Auto-detect</string>`. 

The text SYS_CSVIMPORT_ENC_DQUOTE and some other should not be a translation string. It should be a preference or we set it fix in the code if we have no preference for it.

kainhofer:
I have now made the translatable field identifier more generic. The DQUOTE and QUOTE fields should be translatable, because it makes sense to add an explanation (which I did in my last commit), too.

Fasse:
I saw that you use composer for the external libs and therefore create the folder libs/server/vendor . Since the folder structure is very deep I think if we should add a vendor folder within the adm_program folder. e.g. `adm_program/vendor` instead of `adm_program/libs/server/vendor` or should it be something like `adm_program/vendor_server`

What do you think?

adm_program/languages/en.xml
```diff
@@ -293,6 +293,7 @@
   <string name="MEM_IMPORT">Import</string>
   <string name="MEM_IMPORT_SUCCESSFUL">Import was successful..\n\n#VAR1# new users have been created.\n#VAR2# users have been edited.\n#VAR3# role memberships have been assigned.</string>
   <string name="MEM_IMPORT_USERS">Import user(s)</string>
+  <string name="MEM_IMPORT_SHEETNOTEXISTS">The worksheet #VAR1# does not exist in the imported file!</string>
```
Fasse, line 296: Better name MEM_IMPORT_SHEET_NOT_EXISTS

adm_program/languages/en.xml
```diff
@@ -752,6 +754,18 @@
...
   <string name="SYS_CREATION_DATE">Created on</string>
   <string name="SYS_CSV">CSV</string>
+  <string name="SYS_AUTO_DETECT">Auto-detect</string>
+  <string name="SYS_COMMA">Comma (,)</string>
+  <string name="SYS_SEMICOLON">Semicolon (;)</string>
+  <string name="SYS_TAB">Tab</string>
+  <string name="SYS_PIPE">Pipe (|)</string>
+  <string name="SYS_FIELDSEPARATOR">Separator for CSV-file</string>
+  <string name="SYS_NOQUOTE">No quotation</string>
+  <string name="SYS_DQUOTE">"</string>
```
Fasse, line 764: Duplicate see below.

adm_program/languages/en.xml
```diff
@@ -752,6 +754,18 @@
...
+  <string name="SYS_QUOTE">'</string>
```
Fasse, line 765: Duplicate see below

adm_program/languages/en.xml
```diff
@@ -752,6 +754,18 @@
   <string name="SYS_CREATED_BY">Created by #VAR1# on #VAR2#</string>
   <string name="SYS_CREATION_DATE">Created on</string>
   <string name="SYS_CSV">CSV</string>
+  <string name="SYS_AUTO_DETECT">Auto-detect</string>
+  <string name="SYS_COMMA">Comma (,)</string>
+  <string name="SYS_SEMICOLON">Semicolon (;)</string>
+  <string name="SYS_TAB">Tab</string>
+  <string name="SYS_PIPE">Pipe (|)</string>
+  <string name="SYS_FIELDSEPARATOR">Separator for CSV-file</string>
+  <string name="SYS_NOQUOTE">No quotation</string>
```
Fasse, line 763: Better SYS_NO_QUOTATION

adm_program/languages/en.xml
```diff
@@ -752,6 +754,18 @@
   <string name="SYS_CREATED_BY">Created by #VAR1# on #VAR2#</string>
   <string name="SYS_CREATION_DATE">Created on</string>
   <string name="SYS_CSV">CSV</string>
+  <string name="SYS_AUTO_DETECT">Auto-detect</string>
+  <string name="SYS_COMMA">Comma (,)</string>
+  <string name="SYS_SEMICOLON">Semicolon (;)</string>
+  <string name="SYS_TAB">Tab</string>
+  <string name="SYS_PIPE">Pipe (|)</string>
+  <string name="SYS_FIELDSEPARATOR">Separator for CSV-file</string>
```
Fasse, line 762: Better name SYS_SEPARATOR_FOR_CSV_FILE

adm_program/languages/en.xml
```diff
@@ -862,6 +876,8 @@
   <string name="SYS_ENABLE_DOCUMENTS_FILES_MODULE_DESC">The documents &amp; files module can be activated or deactivated via this setting. (Default: yes)</string>
   <string name="SYS_ENABLE_GROUPS_ROLES_MODULE">Enable groups &amp; roles module</string>
   <string name="SYS_ENABLE_GROUPS_ROLES_MODULE_DESC">Displays the groups and roles of the organization and allows free configuration of groups and roles. Own list can be configured and applied. This module is generally only available for logged in users. (Default: yes)</string>
+  <string name="SYS_ENCODING_DEFAULT">Default encoding (UTF-8)</string>
```
Fasse, line 879: Better SYS_DEFAULT_ENCODING_UTF8

adm_program/languages/en.xml
```diff
@@ -1403,4 +1423,10 @@
   <string name="SYS_YOUR_LISTS">Own lists</string>
   <string name="SYS_YOUR_NAME">Your name</string>
   <string name="SYS_Z_TO_A">Z to A</string>
+  <string name="SYS_FILE_XLSX">Excel 2007-365 (*.xlsx, *.xlsm, *.xlsb)</string>
```
Fasse, line 1426: SYS_EXCEL_2007_365

adm_program/languages/en.xml
```diff
@@ -1403,4 +1423,10 @@
...
+  <string name="SYS_FILE_XLS">Excel 97-2003 (*.xls, *.xlm)</string>
```
Fasse, line 1427: SYS_EXCEL_97_2003

adm_program/languages/en.xml
```diff
@@ -1403,4 +1423,10 @@
...
+  <string name="SYS_FILE_ODS">ODF Spreadsheet (*.ods)</string>
```
Fasse, line 1428: SYS_ODF_SPREADSHEET

adm_program/languages/en.xml
```diff
@@ -1403,4 +1423,10 @@
...
+  <string name="SYS_FILE_CSV">Comma-/tab-separated file (*.csv)</string>
```
Fasse, line 1429: ...name="SYS_COMMA_SEPARATED_FILE">Comma-separated file (*.csv)</string>

adm_program/languages/en.xml
```diff
@@ -1403,4 +1423,10 @@
...
+  <string name="SYS_FILE_HTML">HTML table</string>
```
Fasse, line 1430: SYS_HTML_TABLE

Fasse (changes requested):

kainhofer:
I have now properly renamed all translatable string identifiers per your convention and placed them in the corresponding (alphabetic) position in the language files. I have also squashed all commits into two (one purely for importing the phpSpreadsheet library and the other with the actual changes to admidio using the phpSpreadsheet library).

Fasse (changes requested):
Ok, the code looks fine.

I found one thing when testing the new feature:
- explicit select csv file format
- select a csv file
- add a role to which the members should be added
- click "Weiter"

At the following page this error is shown: 
> Your requested sheet index: 1 is out of bounds. The actual number of sheets is 1.

PR #1044, opened 2021-05-15
Profile fields: Implement drag'n'drop of profile fields using jQuery sortable

kainhofer:


adm_program/languages/en.xml
```diff
@@ -1175,6 +1175,7 @@
   <string name="SYS_REMEMBER_ME">Stay signed in</string>
   <string name="SYS_REMOVE">Remove</string>
   <string name="SYS_REMOVE_MESSAGE">Remove message</string>
+  <string name="SYS_REORDER">Move #VAR1#</string>
```
Fasse, line 1178: We should use the english translation for the string name. So you know within the code better what you mean. Here SYS_MOVE_VAR would be a better name.

adm_program/modules/profile-fields/profile_fields.php
```diff
@@ -193,7 +205,10 @@
             '<i class="fas fa-chevron-circle-up" data-toggle="tooltip" title="' . $gL10n->get('SYS_MOVE_UP', array('MEM_PROFILE_FIELD')) . '"></i></a>
         <a class="admidio-icon-link" href="javascript:void(0)" onclick="moveTableRow(\''.TableUserField::MOVE_DOWN.'\', \'row_usf_'.$usfId.'\',
             \''.SecurityUtils::encodeUrl(ADMIDIO_URL . FOLDER_MODULES . '/profile-fields/profile_fields_function.php', array('mode' => 4, 'usf_id' => $usfId, 'sequence' => TableUserField::MOVE_DOWN)) . '\')">'.
-            '<i class="fas fa-chevron-circle-down" data-toggle="tooltip" title="' . $gL10n->get('SYS_MOVE_DOWN', array('MEM_PROFILE_FIELD')) . '"></i></a>',
+            '<i class="fas fa-chevron-circle-down" data-toggle="tooltip" title="' . $gL10n->get('SYS_MOVE_DOWN', array('MEM_PROFILE_FIELD')) . '"></i></a>
+        <a class="admidio-icon-link">'.
+            '<i class="fas fa-arrows-alt handle" title="' . $gL10n->get('SYS_REORDER', array('MEM_PROFILE_FIELD')) . '"></i></a>
```
Fasse, line 210: We should add data-toggle="tooltip" so the layout of the tooltip fit the other tooltips.

Fasse (changes requested):

Fasse:
Nice feature. I'm not sure if we should use this as the only feature to move rows. How does it work on mobile?

Fasse:
I tested it on mobile and it doesn't work there, so we still need both solutions. Maybe we could hide and show only one solution depending if its a touch device?
https://stackoverflow.com/questions/4817029/whats-the-best-way-to-detect-a-touch-screen-device-using-javascript/4819886#4819886

Fasse:
@kainhofer Could you implement my requested changes, than we can merge this.

PR #1068, opened 2021-07-24
Issue 1067: Implement admin notification mails for all user/profile/membership changes and user creations/deletions

### LINKED ISSUE #1067: System Notification for user profile field changes (and role changes)
In our association, we would like to have a notification mail to the administrator, whenever a user (or an admin) changes a profile field, so we can monitor the data quality.
Also, if a user is deleted or created, a similar notification should be sent.
Also, if a role/group membership is changed/added/deleted, a notification should be sent.

kainhofer:
Notifications for all changes to a user, their profile fields and memberships are implemented via a dedicated class ChangeNotification, which queues all changes and sends out the mail notifications (if system notifications are enabled in the configuration) when the PHP script is done processing the whole page (but just before all global objects are destroyed).
A global object gChangeNotifications is automatically created and all changes in the User and TableMembers classes are properly logged to this object. If any plugin or code directly modifies the database, it needs to use any of the member functions of the ChangeNotification class to register the change.
    
In the past, all profile field changes were logged to the database already. This functionality has now been moved to the ChangeNotification class, so this class is the control center for all change notifications and logging. Implementing history log entries for user creation/deletion, password/username changes or membership changes can be easily added in the corresponding functions in the ChangeNotification class.

Fixes #1067

### DIFF
--- a/adm_program/installation/db_scripts/preferences.php
+++ b/adm_program/installation/db_scripts/preferences.php
@@ -62,6 +62,7 @@ $defaultOrgPreferences = array(
     'enable_system_mails'         => '1',
     'email_administrator'         => 'administrator@'. DOMAIN,
     'enable_email_notification'   => '0',
+    'enable_email_changenotification'   => '0',
 
     // Captcha
     'captcha_type'                => 'pic',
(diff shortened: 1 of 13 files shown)


Fasse:
Thanks for this implementation. That sounds very useful. I will have a Look at it later this week.

Fasse:
Thanks for this great enhancements!

PR #1072, opened 2021-07-25
Composer: use the vanilla/htmlawed package instead of htmlawed/htmlawed, since Htmlawed::filter is not available otherwise

kainhofer:
The admidio code uses Htmlawed::filter in two places, but the object-oriented Htmlawed class is not available in the htmlawed/htmlawed package (=> "Fatal error: Uncaught Error: Class 'Htmlawed' not found in [...]" whenever editing profile field defintion that has its description field not empty.). 

So I simply switched composer.json to use vanilla/htmlawed instead and ran "composer update htmlawed/htmlawed vanilla/htmlawed" to set up the vanilla package instead. 

The only manual change is in composer.json, the rest is auto-generated by compuser update.

With this patch, it is finally possible to edit profile field definitions and give a description for the profile field.


Fasse:
Thanks for the fix.

PR #1078, opened 2021-07-31
Fix password hashing (check for hashing need used other options than password hashing itself)

kainhofer:
Without this patch, the check whether the password needs to be re-hashed on login used the default cost of 10, while the subsequent password hashing used the calculated cost of 11, so the password was re-hashed on each login. This patch makes sure that the same cost option is used for both the hash as well as the rehashing check.


Fasse:
Thank you for fixing this. I will merge this to version 4.0.8 too.

PR #1089, opened 2021-08-11
Fix profile field internal names (remove special chars, properly enclose all SQL field names with ticks in ListConfiguration::getSQL); escape input field values

### LINKED ISSUE #1088: Internal field names with hyphens break SQL
**Describe the bug**
When an internal profile field name contains a hypen (e.g. E-MAIL_ALTERNATIVE), the SQL created to display a list containing that field is invalid and causes a 500 error.

**To Reproduce**
1. Create new profile field named "E-Mail alternative" => the internal name will be 'E-MAIL_ALTERNATIVE'
2. Edit a list (or create a new list) that contains this new profile field
3. Display the newly created list containing that field

There are actually two issues involved

- [ ] Newly created profile fields will not remove hyphens from their internal name (spaces and other special characters are removed)
- [ ] The SQL code in ListConfiguration::getSQL does not properly enclose field names with `....`. Enclosing fields with ticks is highly recommended and allows special characters (including spaces, hyphens etc.) in SQL field name.

The generic issue, of course, is that most queries in Admidio are constructed directly as SQL strings, rather than going through some kind of library that properly escapes fields and special characters (e.g. additional protection from SQL injection attacks and from invalid SQL code due to special characters).

kainhofer:
1) If an internal profile field name contains a hyphen, ListConfiguraiton::getSQL creates invalid SQL, because it does not enclose all field names with ticks. Hyphens in SQL field names are only allowed if the field name is quoted with ticks.

This patch fixed the ListConfiguration::getSQL method to add the ticks, but the issue potentially also appears in other places.

2) When a new profile field is created, this patch also removes all special characters except Letters, Digits and Underscores (spaces are converted to underscore).

3) Input field values are written as html attributes, so their values need to be escapes before inserting it into html. E.g. if a name contains a double quote, invalid HTML will be generated and the field value will be truncated. With the htmlspecialchars, all special characters are properly escaped so that the displayed input field shows the correct (special) characters.

Fixes #1088

### DIFF
--- a/adm_program/system/classes/HtmlElement.php
+++ b/adm_program/system/classes/HtmlElement.php
@@ -506,7 +506,7 @@ abstract class HtmlElement
         $attributes = array();
         foreach ($elementAttributes as $key => $value)
         {
-            $attributes[] = $key . '="' . $value . '"';
+            $attributes[] = $key . '="' . htmlspecialchars($value) . '"';
         }
 
         return ' ' . implode(' ', $attributes);
--- a/adm_program/system/classes/TableUserField.php
+++ b/adm_program/system/classes/TableUserField.php
@@ -126,7 +126,7 @@ class TableUserField extends TableAccess
      */
     private function getNewNameIntern($name, $index)
     {
-        $newNameIntern = strtoupper(str_replace(' ', '_', $name));
+        $newNameIntern = strtoupper(preg_replace('/[^A-Za-z0-9_]/', '', str_replace(' ', '_', $name)));
 
         if ($index > 1)
         {


Fasse:
Hmm the problem with the ticks will be with PostgreSQL. There this SQL syntax won't work :(
Maybe we should disallow hyphen within the internal field name and convert it to `_`?

kainhofer:
> Hmm the problem with the ticks will be with PostgreSQL. There this SQL syntax won't work :(
> Maybe we should disallow hyphen within the internal field name and convert it to `_`?
> 
> So I would like to revert the changes within ListConfiguration.php and only commit the other changes.

Sure, that will prevent the problem in the future. Existing profile fields with hyphens and other special chars will need to be manually fixed in the database (easy to do, since the internal profile field names are only stored in the profile fields table, but not anywhere else).

Fasse:
yes I know that the problem for existing fields will after the remove still exists. But the other syntax breaks PostgreSQL.

Maybe we add a update sql that will fix some of the invalid char in usf_name_intern.

Could you please revert the changes within ListConfiguration.php .

kainhofer:
Okay, did a rebase that dropped the changes to ListConfiguration.php, so only the internal name cleaning and the htmlspecialchars escaping is left in this fix.

Fasse:
Thanks!

PR #1170, opened 2022-01-16
Demo Data: Add table names to translations to prevent errors

### LINKED ISSUE #1169: Demo Data cannot be installed
**Describe the bug**
I set up a clean instance of mysql/mariadb and apache and tried to insert the demo data via call to "http://localhost:81/demo_data/build.php?lang=en"

However, this lead to a SQL-Error:
```
Start installing ...
Reading file "data.sql" ...
Read file "data.sql" finished!
Executing "data.sql" SQL-Statements ...

S Q L - E R R O R

CODE: 42000

1064

You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'VALUES (10, 'SYSTEM', 'Admidio Core', 'CORE', '4.0.7', 0, 730) ' at line 6

B A C K T R A C E

FILE: adm_program/system/classes/Database.php
LINE: 720
CALL: Database->showError()

[...]
```

After some debugging, it turns out that the install script replaces all %PREFIX% by DEMO, so the SQL code is 
```{sql}
INSERT INTO %PREFIX%_components (com_id, com_type, com_name, com_name_intern, com_version, com_beta, com_update_step)
                         VALUES (10, 'SYSTEM', 'Admidio Core', 'CORE', '4.0.7', 0, 730)
```
On the other hand, the demo_data/build.php script does a preg_match_all on (DEMO_\w*) and runs the result though $gL10n->get, i.e. it tries to translate the table name DEMO_components to either German or English.

Unfortunately, the language files do not contain translations for the tables, so the Language::get function (adm_program/system/classes/Language.php, line 122) throws an exception and wraps all untranslated strings in "#". Thus, the SQL turns into 
```{sql}
INSERT INTO #DEMO_components# (com_id, com_type, com_name, com_name_intern, com_version, com_beta, com_update_step)
                         VALUES (10, 'SYSTEM', 'Admidio Core', 'CORE', '4.0.7', 0, 730)
```

Clearly, the table name should not be translated, but the demo data install script treats all strings starting with DEMO_ as translatable... 

The easiest way to solve the issue is to simply add all table names to the languange files, so no exception is thrown. The other approach would be in build.php to explicitly check for all known table names and not to dall $gL10n->get on the table names...

I cannot think of a generic approach to prevent table names from being translated (or to prevent $gL10n->get from wrapping untranslated table names with #)...

kainhofer:
The demo data install script tries to translate all strings starting with DEMO_...
Unfortunately, all database table names also start with DEMO_, so the script tries
to translate them, but since they cannot be found in the language files, Language::get
wraps them in hash marks, producing invalid SQL.

The easiest fix is to simply add all demo table names to the language files.

Fixes and closes #1169

### DIFF
--- a/demo_data/languages/de.xml
+++ b/demo_data/languages/de.xml
@@ -72,4 +72,37 @@
     <string name="DEMO_ROOM_MEETING_ROOM_CONTENT">In diesem Raum können Besprechungen stattfinden. Der Raum muss vorher reserviert werden. Ein Beamer steht zur Verfügung.</string>
     <string name="DEMO_ROOM_FUNCTION_ROOM">Veranstaltungsraum</string>
     <string name="DEMO_ROOM_FUNCTION_ROOM_CONTENT">Der Veranstaltungsraum kann für Geburtstagsfeiern, Jahresversammlungen oder Party's genutzt werden. Eine rechtzeitige Reservierung ist erwünscht.</string>
+
+    <string name="DEMO_components">DEMO_components</string>
+    <string name="DEMO_organizations">DEMO_organizations</string>
+    <string name="DEMO_menu">DEMO_menu</string>
+    <string name="DEMO_preferences">DEMO_preferences</string>
+    <string name="DEMO_users">DEMO_users</string>
+    <string name="DEMO_categories">DEMO_categories</string>
+    <string name="DEMO_category_report">DEMO_category_report</string>
+    <string name="DEMO_roles">DEMO_roles</string>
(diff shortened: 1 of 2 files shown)


Fasse:
It's strange, I don't get that error when inserting the data in my Apache / Mysql database.

But I think it would be a better solution to change the prefix of the translation strings within the demo_data folder, so it's not DEMO maybe DDT (demo data). After that we don't have the same prefix like the table prefix.

Fasse:
@kainhofer I have made a fix that is referenced in #1169 . 
Please try this with your installation.

PR #1205, opened 2022-02-04
Update README-Docker.md

maeries:
- fixed typo contaner_name -> container_name
- changed volume path on the host side. alternatively a volume declaration could have been added
- changed admidio environment variable notation to be uniform with the notation used for the db
- removed the ports from the database. you don't need to open them as both services are in the same network


Fasse:
@maeries Thanks for the changes.

PR #1247, opened 2022-03-20
Scan for vulnerabilities in Admidio Docker Images with trivy

schast:
* hooks/build:
  * add ADMIDIO_TRIVY_ENABLED to build args
* Dockerfile:
  * delete trivy cache after testing
  * delete unused, old and unsecure nodejs version from image


Fasse:
Could you please commit this changes also to the `master`, so we have it in future versions too.

PR #1282, opened 2022-06-11
Update photos.php

frest-de:
https://www.admidio.org/forum/viewtopic.php?f=3&t=9260
The new album should not use the uuid of another album, otherwise it would get overridden


Fasse:
I have fix the problem with this issue #1283 

There were several problems with the given uuid when creating a new album. So I will deny this pull request because it has the wrong fix.

PR #1286, opened 2022-06-17
Fixed docker environment variables not used

ChaseMB00:
I had some problems running Admidio behind a reverse proxy, as my environment variables had no effect. So I went digging and found, that the Admidio config is actually not generated. I changed the script to check if necessary environment variables are set and generate the config.


Fasse:
@ChaseMB00 thanks for this enhancement to the script.

ChaseMB00:
No problem, I had fun figuring out what went wrong.

PR #1297, opened 2022-08-13
Receipient Placeholder in email wrapper

marchorst:
With this change you are able to use the following placeholders within the email wrapper.
'#receiver_first_name#'
'#receiver_firstname#'
'#receiver_surname#'
'#receiver_lastname#'
'#receiver_email#'
'#receiver_name#'

It is also prepared to add more fields

### DIFF
--- a/adm_program/languages/de-DE.xml
+++ b/adm_program/languages/de-DE.xml
@@ -1043,6 +1043,10 @@
   <string name="SYS_MOVE_VAR">#VAR1# verschieben</string>
   <string name="SYS_MULTIPLE_RECIPIENTS">Mehrere Empfänger:innen</string>
   <string name="SYS_MULTIPLE_RECIPIENTS_DESC">E-Mails an mehrere Empfänger:innen enthalten alle Empfänger:innen im BCC-Feld. Damit ist sichergestellt, dass keiner die E-Mail-Adressen der anderen Empfänger:innen sieht. Das TO-Feld muss allerdings auch befüllt werden. Über diese Einstellung kann festgelegt werden, was im Empfängerfeld (TO) hinterlegt wird. Dies kein ein versteckter Wert \"Undisclosed Recipients\" sein, was aber bei manchen Providern zu Problemen führt. Alternativ kann das Empfängerfeld (TO) mit der Absenderadresse bzw. der Administrator:innen-Adresse aus den Systembenachrichtigungen gefüllt werden. Hinweis: Alle Emails an mehrere Empfänger:innen werden dann auch an die E-Mail-Adresse im Empfängerfeld (TO) geschickt. (Voreinstellung: Absender:in)</string>
+  <string name="SYS_MAIL_SENDING_MODE">E-mail Versand</string>
+  <string name="SYS_MAIL_SENDING_MODE_DESC">Im Modus "Gebündelter versandt" werden E-Mails in einer oder mehrerer gebündelten E-Mails versendet. Dabei werden die Empfänger als BCC oder TO eingefügt. Im Modus "Einzelempfänger" wird jede E-Mail einzeln an den entsprechenden Empfänger im TO Feld versendet. In diesem Modus gibt es weitere Platzhalter die verwendet werden können:\nEmpfänger Vorname:\n#receiver_first_name# #receiver_firstname# #receiver_surname#\nEmpfänger Nachname:\n#receiver_lastname#\nEmpfänger E-Mail:\n#receiver_email#\nEmpfänger Name:\n#receiver_name#</string>
+  <string name="SYS_MAIL_BULK">Gebündelter versandt</string>
+  <string name="SYS_MAIL_SINGLE">Einzelempfänger</string>
   <string name="SYS_MUST_HAVE_ADMINISTRATOR">Die Rolle Administrator:in muss mindestens ein Mitglied haben!</string>
   <string name="SYS_MYLIST_CONDITION_DESC">Hier können Sie Bedingungen zu jedem Feld in Ihrer Liste hinterlegen. Damit werden die ermittelten Mitglieder der ausgewählten Rolle mit Ihren Bedingungen weiter eingeschränkt.</string>
   <string name="SYS_NAME">Name</string>
--- a/adm_program/languages/de-DE.xml
+++ b/adm_program/languages/de-DE.xml
@@ -1043,10 +1043,10 @@
   <string name="SYS_MOVE_VAR">#VAR1# verschieben</string>
   <string name="SYS_MULTIPLE_RECIPIENTS">Mehrere Empfänger:innen</string>
   <string name="SYS_MULTIPLE_RECIPIENTS_DESC">E-Mails an mehrere Empfänger:innen enthalten alle Empfänger:innen im BCC-Feld. Damit ist sichergestellt, dass keiner die E-Mail-Adressen der anderen Empfänger:innen sieht. Das TO-Feld muss allerdings auch befüllt werden. Über diese Einstellung kann festgelegt werden, was im Empfängerfeld (TO) hinterlegt wird. Dies kein ein versteckter Wert \"Undisclosed Recipients\" sein, was aber bei manchen Providern zu Problemen führt. Alternativ kann das Empfängerfeld (TO) mit der Absenderadresse bzw. der Administrator:innen-Adresse aus den Systembenachrichtigungen gefüllt werden. Hinweis: Alle Emails an mehrere Empfänger:innen werden dann auch an die E-Mail-Adresse im Empfängerfeld (TO) geschickt. (Voreinstellung: Absender:in)</string>
-  <string name="SYS_MAIL_SENDING_MODE">E-mail Versand</string>
-  <string name="SYS_MAIL_SENDING_MODE_DESC">Im Modus "Gebündelter versandt" werden E-Mails in einer oder mehrerer gebündelten E-Mails versendet. Dabei werden die Empfänger als BCC oder TO eingefügt. Im Modus "Einzelempfänger" wird jede E-Mail einzeln an den entsprechenden Empfänger im TO Feld versendet. In diesem Modus gibt es weitere Platzhalter die verwendet werden können:\nEmpfänger Vorname:\n#receiver_first_name# #receiver_firstname# #receiver_surname#\nEmpfänger Nachname:\n#receiver_lastname#\nEmpfänger E-Mail:\n#receiver_email#\nEmpfänger Name:\n#receiver_name#</string>
-  <string name="SYS_MAIL_BULK">Gebündelter versandt</string>
-  <string name="SYS_MAIL_SINGLE">Einzelempfänger</string>
+  <string name="SYS_MAIL_SENDING_MODE">Versandmodus</string>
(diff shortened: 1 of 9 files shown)


Fasse:
Could you please explain, why you remove the preferences with the option to add the recipients to BCC and set the count for it?

marchorst:
Sure, because we are sending ever email on its one, the receiver will always be alone in the "TO" field, so no need to move them to BCC. Hope this makes sense?

Fasse:
But we need this preferences. Some hosters don't allow to send many emails in a short time slot. Therefore we have the preferences to send one email to several recipients.

That was the problem why we don't had these placeholders. But if we want these placeholders, we could only use them if we send one email to one recipient. Maybe these placeholders will only be used if the preference is set to 1.

Fasse (changes requested):
But we need this preferences. Some hosters don't allow to send many emails in a short time slot. Therefore we have the preferences to send one email to several recipients.

That was the problem why we don't had these placeholders. But if we want these placeholders, we could only use them if we send one email to one recipient. Maybe these placeholders will only be used if the preference is set to 1.

marchorst:
> But we need this preferences. Some hosters don't allow to send many emails in a short time slot. Therefore we have the preferences to send one email to several recipients.
> 
> That was the problem why we don't had these placeholders. But if we want these placeholders, we could only use them if we send one email to one recipient. Maybe these placeholders will only be used if the preference is set to 1.

Makes sense to me.
Should we add another option like "Sending Method: <Single, Bulk>" and hide the options "Multiple recipients" and "Number of recipients" if it is set to "Single"?

Fasse:
Hmm, that could be an option. But if you don't want to add a new preference it should also work for me if the check is for preference set to 1. Than we could also set the default for new installations to 1 or to single.

marchorst:
I have now added it to the preferences.
I still need to adjust the language fields a bit, but the logic itself should work as expected.

I would say it makes more sense to add it to the preferences and make a good description, because the "Multiple Recipients" and "Number of Recipients" have no effect after that.

adm_program/modules/preferences/preferences.php
```diff
@@ -513,25 +513,38 @@ function getPreferencePanel($group, $id, $parentId, $title, $icon, $body)
...
-    $gL10n->get('SYS_NUMBER_RECIPIENTS'),
-    $formValues['mail_number_recipients'],
-    array('type' => 'number', 'minNumber' => 0, 'maxNumber' => 9999, 'step' => 1, 'helpTextIdInline' => 'SYS_NUMBER_RECIPIENTS_DESC')
+    array('defaultValue' => $formValues['mail_sending_mode'], 'showContextDependentFirstEntry' => false, 'helpTextIdInline' => 'SYS_MAIL_SENDING_MODE_DESC')
 );
+
+$sendingMode = $gSettingsManager->getInt('mail_sending_mode');
+
+// Only show this options if SendingMode = 0 / SYS_MAIL_BULK
+if($sendingMode == 0) {
```
Fasse, line 527: This wont work in the form because you check the mode within php. If you want to show or hide those fields this must be done with javascript dynamically if the user select each mode. Otherwise the prefences should be shown always.

adm_program/system/classes/Email.php
```diff
@@ -150,18 +158,22 @@ public function __construct()
      * in the email send process.
      * @param string $address A valid email address to which the email should be sent.
      * @param string $name    The name of the recipient that will be shown in the email header.
+     * @param array $additionalFields    Additional fields to map in a Key Value like Array. Not used at all yet.
      * @return bool Returns **true** if the address was added to the recipients list.
      */
-    public function addRecipient($address, $name = '')
+    public function addRecipient($address, $firstName = '', $name = '', $additionalFields = array())
```
Fasse, line 164: Have you test you changes? You change the parameters of the method but don't change the calls of the method. Also currently the "name" is "Firstname Lastname" and not only the lastname.

adm_program/system/classes/Email.php
```diff
@@ -565,135 +577,137 @@ public function setTemplateText($text, $senderName, $senderEmail, $senderUuid, $
...
     }
 
     /**
-     * Sends a copy of the mail back to the sender. If the flag emListRecipients it set than all
-     * recipients will be listed in the mail.
-     * @throws \PHPMailer\PHPMailer\Exception
+     * Funktion um den Nachrichtentext an die Mail uebergeben
+     * @param string $text
      */
-    private function sendCopyMail()
```
Fasse, line 586: Why did you remove the sendCopyMail method? As I see there is no alternativ implementation of this feature. Within the copy mail you got the information to which roles the mail was sent.

Fasse (approved):

adm_program/modules/preferences/preferences.php
Fasse, line 527: This wont work in the form because you check the mode within php. If you want to show or hide those fields this must be done with javascript dynamically if the user select each mode. Otherwise the prefences should be shown always.

adm_program/system/classes/Email.php
Fasse, line 164: Have you test you changes? You change the parameters of the method but don't change the calls of the method. Also currently the "name" is "Firstname Lastname" and not only the lastname.

This belongs also to the other methods.

adm_program/system/classes/Email.php
Fasse, line 586: Why did you remove the sendCopyMail method? As I see there is no alternativ implementation of this feature. Within the copy mail you got the information to which roles the mail was sent.

Fasse (changes requested):

adm_program/system/classes/Email.php
marchorst: Should be fixed within the next commit.

adm_program/system/classes/Email.php
marchorst: This was deleted unintentionally, I have restored it.

adm_program/modules/preferences/preferences.php
```diff
@@ -513,6 +513,24 @@ function getPreferencePanel($group, $id, $parentId, $title, $icon, $body)
     $formValues['mail_sendmail_name'],
     array('maxLength' => 50, 'helpTextIdInline' => 'SYS_SENDER_NAME_DESC')
 );
+
+// Add js to show or hide mail options
+$page->addJavascript('
```
marchorst, line 518: I am not sure if this is the best place to add it or if there is any "default" js how to handle things like this.

adm_program/languages/de-DE.xml
```diff
@@ -1043,6 +1043,10 @@
   <string name="SYS_MOVE_VAR">#VAR1# verschieben</string>
   <string name="SYS_MULTIPLE_RECIPIENTS">Mehrere Empfänger:innen</string>
   <string name="SYS_MULTIPLE_RECIPIENTS_DESC">E-Mails an mehrere Empfänger:innen enthalten alle Empfänger:innen im BCC-Feld. Damit ist sichergestellt, dass keiner die E-Mail-Adressen der anderen Empfänger:innen sieht. Das TO-Feld muss allerdings auch befüllt werden. Über diese Einstellung kann festgelegt werden, was im Empfängerfeld (TO) hinterlegt wird. Dies kein ein versteckter Wert \"Undisclosed Recipients\" sein, was aber bei manchen Providern zu Problemen führt. Alternativ kann das Empfängerfeld (TO) mit der Absenderadresse bzw. der Administrator:innen-Adresse aus den Systembenachrichtigungen gefüllt werden. Hinweis: Alle Emails an mehrere Empfänger:innen werden dann auch an die E-Mail-Adresse im Empfängerfeld (TO) geschickt. (Voreinstellung: Absender:in)</string>
+  <string name="SYS_MAIL_SENDING_MODE">E-mail Versand</string>
```
Fasse, line 1046: We should name it "Versandmodus" and it the values should be "Gebündelt" and "Einzeln". Within the description we should not mention the parameters. Better we should explain the reason for "Gebündelt". Maybe the description could be something like this "Im Modus 'Gebündelt' werden E-Mails an mehrere Empfänger zusammengefasst. Eine E-Mail erhält dann eine konfigurierte Anzahl an Empfänger im TO oder BCC Feld. Dies ist sinnvoll wenn der Hoster den Versand von vielen E-Mails nicht erlaubt. Im Modus 'Einzeln' erhält jeder Empfänger eine separate E-Mail. In diesem Modus können einer E-Mail-Vorlage weitere Felder hinzugefügt werden."

adm_program/modules/ecards/ecard_function.php
```diff
@@ -156,7 +156,7 @@ public function parseEcardTemplate($imageName, $ecardMessage, $ecardData, $recip
      * @param string $photoServerPath der Pfad wo die Bilder in der Grußkarte am Server liegen
      * @return bool|string
      */
-    public function sendEcard($senderName, $senderEmail, $ecardHtmlData, $recipientName, $recipientEmail, $photoServerPath)
+    public function sendEcard($senderName, $senderEmail, $ecardHtmlData, $recipientFirstName, $recipientName, $recipientEmail, $photoServerPath)
```
Fasse, line 159: Could we use `$recipientLastName` so it's clear that its the lastname.

adm_program/system/classes/Email.php
Fasse: Could we use $lastName instead of $name so it's clear that its the lastname.

adm_program/system/classes/Email.php
```diff
@@ -323,10 +335,10 @@ public function addRecipientsByUserId($userId)
      * @param string $name
      * @return true|string
      */
-    public function addCopy($address, $name = '')
+    public function addCopy($address, $firstName = '', $name = '')
```
Fasse, line 338: Could we use $lastName instead of $name so it's clear that its the lastname.

adm_program/modules/preferences/preferences.php
Fasse: until now we have no default to handle things like this. But a default could be useful :-)

adm_program/system/classes/Email.php
```diff
@@ -346,9 +358,9 @@ public function addCopy($address, $name = '')
      * @return bool
      * @deprecated 4.2.0:4.3.0 "addBlindCopy()" is deprecated, use "addRecipient()" instead.
      */
-    public function addBlindCopy($address, $name = '')
+    public function addBlindCopy($address, $firstName = '', $name = '')
```
Fasse, line 361: Could we use $lastName instead of $name so it's clear that its the lastname.

adm_program/system/classes/Email.php
```diff
@@ -346,9 +358,9 @@ public function addCopy($address, $name = '')
...
     {
-        return $this->addRecipient($address, $name);
+        return $this->addRecipient($address, $firstName, $name);
```
Fasse, line 363: Could we use $lastName instead of $name so it's clear that its the lastname.

adm_program/system/classes/Email.php
```diff
@@ -565,6 +577,32 @@ public function setTemplateText($text, $senderName, $senderEmail, $senderUuid, $
...
+     * @param string $name  Receiver firstname and surname
+     */
+    public function setUserSpecificTemplateText($text, $firstName, $surname, $email, $name)
+    {
+        // replace all line feeds within the mailtext into simple breaks because only those are valid within mails
+        $text = str_replace("\r\n", "\n", $text);
+
+        // replace parameters in email template
+        $replaces = array(
+            '#receiver_first_name#' => $firstName,
```
Fasse, line 596: We should only use one parameter for firstname and one parameter for surname. 
Maybe `#receiver_firstname#` and `#receiver_lastname#` are good ones.

Fasse (changes requested):

Fasse (approved):

Fasse:
@marchorst  Thanks for this great new feature. With the release of version 4.2 I will update the doku with the new available parameters.

xbluemonkx:
> 

Hi, die neuen Parameter fehlen in der Doku leider noch. Musste gerade eine Ewigkeit suchen, um dieses Ticket zu finden: https://www.admidio.org/dokuwiki/doku.php?id=de:2.0:e-mail-templates

Fasse:
This new parameters are not part of the templates. They can be used within the concrete email you want to write. There we have documented them within the email dialog.

Issue #1325, opened 2022-11-12
PHP Fatal error:  Uncaught Error: Call to a member function fetchAll() on bool


WalterWampe:
Error thrown starting up using docker-compose on Ubuntu 22.04.1 LTS.

--- 

## docker-compose.yaml:
```
version: "3.9"


# https://github.com/Admidio/admidio/blob/master/README-Docker.md

services:
  db:
    restart: unless-stopped
    image: mariadb:10.4.13
    container_name: admidio-mariaDB
    volumes:
      - "/opt/volumes/admidio/mariadb/config:/etc/mysql/conf.d"
      - "/opt/volumes/admidio/mariadb/data:/var/lib/mysql"
    # ports:
    #   - 3306:3306
    environment:
      - MYSQL_ROOT_PASSWORD=rootpasswd
      - MYSQL_DATABASE=admidio
      - MYSQL_USER=admidio
      - MYSQL_PASSWORD=password
    healthcheck:
      test: ["CMD", "mysqladmin", "ping", "-u", "root", "--password=rootpasswd"]
      interval: 10s
      timeout: 10s
      retries: 5

  admidio:
    restart: unless-stopped
    image: admidio/admidio:latest
    container_name: admidio
    depends_on:
      # - db
      db:
        condition: service_healthy
    volumes:
      - /opt/volumes/admidio/files:/opt/app-root/src/adm_my_files
      - /opt/volumes/admidio/themes:/opt/app-root/src/adm_themes
      - /opt/volumes/admidio/plugins:/opt/app-root/src/adm_plugins
    ports:
      - 1142:8080
    environment:
      - ADMIDIO_DB_TYPE=mysql
      - ADMIDIO_DB_HOST=db:3306
      - ADMIDIO_DB_NAME=admidio
      - ADMIDIO_DB_USER=admidio
      - ADMIDIO_DB_PASSWORD=password
      - ADMIDIO_DB_TABLE_PRAEFIX=gbv
      # - ADMIDIO_MAIL_RELAYHOST=hostname.domain.at:25
      - ADMIDIO_LOGIN_FOR_UPDATE=1
      - ADMIDIO_ORGANISATION=GBV
      - ADMIDIO_PASSWORD_HASH_ALGORITHM=DEFAULT
      - TZ=Europe/Berlin
      #- HTTPD_START_SERVERS=8
      #- DOCUMENTROOT=/
      #- PHP_MEMORY_LIMIT=256M
[...]

PR #1331, opened 2022-12-24
dockerscripts/startup.sh: add .admidio_installed file to check for fi…

schast:
…rst run, do not generate config.php on first run (#1325)


Fasse:
I would also add it to the master or should it only be for 4.1 ?

schast:
> I would also add it to the master or should it only be for 4.1 ?

@Fasse: yes, please merge to master.

Fasse:
done

PR #1339, opened 2023-01-06
Improve Dockerfile to reduce image size

tdurieux:
Hi there,

I've made a small improvement to the Dockerfile. I added `--no-cache-dir` to the `pip install` command and it reduces the image size by 145.02 MB (6%) without affecting the image.

Impact on the image size:
* Image size before repair: 2.36 GB
* Image size after repair: 2.22 GB
* Difference: 145.02 MB (6%)

I hope that you will find these changes useful to you. :smile: Let me know if you have any questions or concerns.


schast:
The actual image size is about 1.4 GB. There is also no `--no-cache-dir` option added and there is no pip.
The directory `/var/cache/yum` is already removed with `yum -y clean all`.
Therefore, we will close this pull request.

PR #1480, opened 2023-09-17
Rename members module to contacts

Fasse:


Fasse:
Enable contacts module for all registered users #1481

PR #1510, opened 2023-10-30
Update README.md

bhargavshirin:
added full_stop in end


Fasse:
Thanks for the hint, but we are planing to deploy a complete new readme.

Issue #1537, opened 2023-12-19
Can't install fresh admidio instance with traefik reverse proxy on docker


cmellwig:
Fresh installation fails when traefik reverse proxy is in place before installation. The container starts and goes into unhealthy state:

```
[INFO ] set filesystem permissions (chown -R www-data:root .)

[INFO ] configure Listen port in /etc/apache2/ports.conf

[INFO ] configure VirtualHost port in /etc/apache2/sites-available/000-default.conf

[INFO ] configure ServerName in /etc/apache2/sites-available/000-default.conf

[INFO ] configure postfix set inet_protocols to ipv4 since ipv6 is disabled (net.ipv6.conf.all.disable_ipv6=1)

[INFO ] execute postfix start

[INFO ] execute postfix status

[INFO ] execute mailq

Mail queue is empty

[INFO ] create .admidio_installed file (/opt/app-root/src/adm_my_files/.admidio_installed)

[INFO ] run apache with php enabled (apache2-foreground)

AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.20.0.7. Set the 'ServerName' directive globally to suppress this message

AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.20.0.7. Set the 'ServerName' directive globally to suppress this message

[Tue Dec 19 10:34:02.982659 2023] [mpm_prefork:notice] [pid 130] AH00163: Apache/2.4.56 (Debian) PHP/8.2.12 configured -- resuming normal operations

[Tue Dec 19 10:34:02.982739 2023] [core:notice] [pid 130] AH00094: Command line: 'apache2 -D FOREGROUND'

[...]

PR #1585, opened 2024-02-13
Improve DB Installation Check (#1537)

schast:
We have problems with the database check from time to time. Currently it is assumed that the database has already been initialized as soon as a config.php exists.

In most cases, the following error occurs:
```
PHP Fatal error:  Uncaught Error: Call to a member function fetchAll() on bool in /opt/app-root/src/adm_program/system/classes/Database.php:516
Stack trace:
#0 /opt/app-root/src/adm_program/system/classes/Database.php(442): Database->loadTableColumnsProperties()
#1 /opt/app-root/src/adm_program/system/classes/TableAccess.php(654): Database->getTableColumnsProperties()
#2 /opt/app-root/src/adm_program/system/classes/TableAccess.php(147): TableAccess->setColumnsInfos()
#3 /opt/app-root/src/adm_program/system/classes/TableAccess.php(109): TableAccess->clear()
#4 /opt/app-root/src/adm_program/system/classes/Session.php(62): TableAccess->__construct()
#5 /opt/app-root/src/adm_program/system/common.php(91): Session->__construct()
#6 /opt/app-root/src/adm_program/overview.php(18): require_once('...')
#7 {main}\n  thrown in /opt/app-root/src/adm_program/system/classes/Database.php on line 516
  127.0.0.1 "GET /adm_program/overview.php HTTP/1.1" 500 -
```

Therefore, the check should take place directly in the database and, if necessary, redirect to the installation page.

With the new check, the installation in the container can be significantly simplified, as most of the values are already available via the container configuration.

@Fasse: Can you please review the pull request (changes in `adm_program`) and test with a normal installation?

### DIFF
--- a/adm_program/installation/index.php
+++ b/adm_program/installation/index.php
@@ -9,8 +9,24 @@
  ***********************************************************************************************
  */
 
+
+$rootPath = dirname(__DIR__, 2);
+
 // check if installation is necessary
-if (is_file('../../adm_my_files/config.php')) {
+if (is_file($rootPath . '/adm_my_files/config.php')) {
+    // load config and init bootstrapping
+    require_once($rootPath . '/adm_my_files/config.php');
+
+    // check for empty db and redirect to installation wizard
+    try {
+        $gDb = Database::createDatabaseInstance();
+        $gDb->getTableColumns($g_adm_db . '.' . $g_tbl_praefix . '_sessions');
+    } catch (\Throwable $t) {
+        $page = 'installation.php';
+        header('Location: ' . $page);
+        exit();
+    }
+
     $page = 'update.php';
 } else {
     $page = 'installation.php';
--- a/adm_program/system/common.php
+++ b/adm_program/system/common.php
@@ -38,6 +38,15 @@ try {
     // => EXIT
 }
 
+// check for empty db and redirect to installation wizard
+try {
+    $gDb->getTableColumns($g_adm_db . '.' . $g_tbl_praefix . '_sessions');
+} catch (\Throwable $t) {
+    header('Location: adm_program/installation/index.php');
+    exit();
+}
+
+
 /*********************************************************************************
  Create and validate sessions, check auto login, read session variables
 /********************************************************************************/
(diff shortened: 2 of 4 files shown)


schast:
current installation steps with Docker Container (6 steps with manually filled db options):

admidio_installation_orig_step01_installation
admidio_installation_orig_step02_connect_database
admidio_installation_orig_step03_create_organisazion
admidio_installation_orig_step04_create_administrator
admidio_installation_orig_step05_create_config
admidio_installation_orig_step06_start_installation

schast:
new installation steps with Docker Container (3 steps with automatically filled by docker env variables).

With this changes the other steps are automatically skipped since all options are already set in config file.

admidio_installation_step01_create_organization
admidio_installation_step02_create_administrator
admidio_installation_step03_start_installation

adm_program/system/common.php
```diff
@@ -38,6 +38,15 @@
     // => EXIT
 }
 
+// check for empty db and redirect to installation wizard
+try {
```
Fasse, line 42: Do we want to check for an empty db in every script? Why not check it only in index.php ?

Fasse (changes requested):

adm_program/system/common.php
schast: because in `adm_program/system/common.php` on Line 35 we have already the `config.php` loaded and opened the db connection.

adm_program/installation/index.php
```diff
@@ -9,8 +9,24 @@
  ***********************************************************************************************
  */
 
+
+$rootPath = dirname(__DIR__, 2);
+
 // check if installation is necessary
-if (is_file('../../adm_my_files/config.php')) {
+if (is_file($rootPath . '/adm_my_files/config.php')) {
```
schast, line 16: the second check here is required to make a difference between normal installation and docker installation.
With a Docker installation, it is possible that a config.php exists, but the database has not yet been initialized.

Fasse (approved):
Classic installation is still working.

PR #1591, opened 2024-02-21
Migrate from vCard 2.1 to vCard 3.0 with utf-8

schast:
- [x] migrate format to vCard 3.0 to support import in modern systems like nextcloud
- [x] switch from iso-8859-1 to utf-8
- [x] only export existing fields
- [x] add REV and UID field to support vcard merging (do not import multiple)

### DIFF
--- a/adm_program/modules/groups-roles/groups_roles_function.php
+++ b/adm_program/modules/groups-roles/groups_roles_function.php
@@ -290,7 +290,7 @@ try {
 
         $filename = FileSystemUtils::getSanitizedPathEntry($filename);
 
-        header('Content-Type: text/x-vcard; charset=iso-8859-1');
+        header('Content-Type: text/vcard; charset=utf-8');
         header('Content-Disposition: attachment; filename="' . $filename . '"');
 
         // necessary for IE, because without it the download with SSL has problems
--- a/adm_program/system/classes/User.php
+++ b/adm_program/system/classes/User.php
@@ -1218,9 +1218,7 @@ class User extends TableAccess
         if ($this->getValue('usr_timestamp_change') !== '') {
             $vCard[] = 'REV:' . $this->getValue('usr_timestamp_change', 'c');
         }
-        if ($this->getValue('usr_uuid') !== '') {
-            $vCard[] = 'UID:urn:uuid:' . $this->getValue('usr_uuid');
-        }
+        $vCard[] = 'UID:urn:uuid:' . $this->getValue('usr_uuid');
         $vCard[] = 'END:VCARD';
         return implode("\r\n", $vCard) . "\r\n";
     }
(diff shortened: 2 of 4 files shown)


adm_program/modules/groups-roles/groups_roles_function.php
```diff
@@ -290,7 +290,7 @@
 
         $filename = FileSystemUtils::getSanitizedPathEntry($filename);
 
-        header('Content-Type: text/x-vcard; charset=iso-8859-1');
```
Fasse, line 293: As I understand https://de.wikipedia.org/wiki/VCard the content-type text/x-vcard should be still used for vcard version 3.0

adm_program/system/classes/User.php
```diff
@@ -1119,93 +1119,109 @@ public function getValue(string $columnName, string $format = '')
...
             }
-
-            $vCard[] = 'X-GENDER:' . $xGender;
-            $vCard[] = 'X-WAB-GENDER:' . $xWabGender;
         }
         if ($this->getValue('usr_timestamp_change') !== '') {
-            $vCard[] = 'REV:' . $this->getValue('usr_timestamp_change', 'Ymd\This');
+            $vCard[] = 'REV:' . $this->getValue('usr_timestamp_change', 'c');
+        }
+        if ($this->getValue('usr_uuid') !== '') {
```
Fasse, line 1221: Every user record should have a usr_uuid so you don't need to check if it's empty or not.

Fasse (commented):
Nice enhancement!

Fasse (changes requested):
Please view my comments.

adm_program/system/classes/User.php
```diff
@@ -1119,93 +1119,109 @@ public function getValue(string $columnName, string $format = '')
...
+            } elseif ((int)$this->getValue('GENDER') === 2) {
+                $vCard[] = 'GENDER:F';
+            } elseif ((int)$this->getValue('GENDER') === 3) {
+                $vCard[] = 'GENDER:O';
             }
-
-            $vCard[] = 'X-GENDER:' . $xGender;
-            $vCard[] = 'X-WAB-GENDER:' . $xWabGender;
         }
         if ($this->getValue('usr_timestamp_change') !== '') {
```
Fasse, line 1218: I don't know exactly how REV worked. But within iCal there is a similar field and that must always be filled. First with usr_timestamp_create and if set than with usr_timestamp_change.

adm_program/system/classes/User.php
schast: thanks, I will fix that

adm_program/system/classes/User.php
schast: ok, REV field is now set always.

Fasse (approved):

PR #1615, opened 2024-03-28
improve line break compatibility for PhpSpreadsheet

Laqoore:
Improvement of the export via PhpSpreadsheet for line breaks in cells for Windows / MacOS / Linux


Fasse:
Thanks for this improvement!

Issue #1617, opened 2024-03-29
Make contacts module search case-insensitive


robertoschwald:
**Is your feature request related to a problem? Please describe.**
In the old Member module, the search was case-insensitive. In the new Contacts module, it is case-sensitive.

**Describe the solution you'd like**
Make the search case-insensitive again, as otherwise, it could be problematic to find the entries.

**Describe alternatives you've considered**
./.

**Additional context**
./.

Fasse:
Within my tests in our demo area, the search in contact module is still case-insensitive.
Could you please provide an example where the search is case sensitive?

robertoschwald:
Here we go:
In Contacts, search for "muster" doesn't show a hit. Searching for "Muster" shows a hit. In Groups&Roles, the search is case-insensitive, but not in Contacts.

<img width="1568" alt="lowercase" src="https://github.com/Admidio/admidio/assets/1519879/bdf47356-3406-4e80-8d7b-64268ade97f6">

<img width="1568" alt="uppercase" src="https://github.com/Admidio/admidio/assets/1519879/af48ef5b-a570-46b9-b558-1ba64fe2e1dc">

robertoschwald:
I'm on MariaDB 10.5. Tables are CHARSET=utf8 COLLATE=utf8_unicode_ci.

In the docker-compose example, you use MariaDB 11.1, where Contacts search is case-insensitive. 

If I change the SQL queries to use explicit LOWER() on the queried fields and the contatenated query, it works independent of the db collation. Maybe thats a way to enhance the code.

robertoschwald:
I added a PR for this . https://github.com/Admidio/admidio/pull/1621

Fasse:
Thanks for the fix. That explains why I could Not reproduce the problem.

PR #1621, opened 2024-04-01
Case-Insensitive searches using Lower()

robertoschwald:
Case-Insensitive search in contacts data independent of db collation


Fasse:
#1617

PR #1683, opened 2024-08-24
Fix for 500 error when null UUID uploaded in user import

beenbiishop:
Since updating to the latest version of Admidio, when I upload CSV files that have some cells blank in the Unique Admidio-ID column using the Import Users tool of the Contacts module, a null value is passed into the readDataByUuid function of the Users class, causing an exception and a subsequent 500 error.

This causes a TypeError, as the signature of this function only allows string values. Here is the exception I get:
`User::readDataByUuid(): Argument #1 ($uuid) must be of type string, null given, called in /my_admidio_root/adm_program/modules/contacts/import_user.php on line 82`

When not importing the UUID column, or when there are no empty cells in the Admidio Unique-ID column of the CSV, no errors happen. Not sure why this only started happening when I switched from using 4.2.4 to 4.3.10, but if I had to guess it probably came from a change to how the files are parsed, or how empty cells are handled in PhpSpreadsheet.


beenbiishop:
Let me know if you want an issue created as well :)

adm_program/modules/contacts/import_user.php
```diff
@@ -79,7 +79,7 @@
 
         $userImport->setImportMode((int)$_SESSION['user_import_mode']);
         if ($identifyUserByUuid) {
-            $userImport->readDataByUuid($line[$importProfileFields['usr_uuid']]);
+            $userImport->readDataByUuid($line[$importProfileFields['usr_uuid']] ?? '');
```
beenbiishop, line 82: a simple replacement of null values with an empty string solved all errors I was encountering :)

Fasse:
Thanks for the fix!

PR #1710, opened 2024-11-29
Add option to use e-Mail-address instead of username for login

### LINKED ISSUE #1071: Alternatively allow e-Mail Adresses for login instead of username
If a user has an account set up (i.e. a login name set), it should be possible to use the e-Mail address instead of the user name for login. Other systems like Nextcloud allow this kind of login, so our users are used to giving their e-Mail instead of the username.

kainhofer:
When the new option (Settings -> Security) is enabled (it's disabled by default), the e-mail-address can be used instead of the username for logging in.

Fixes #1071


Fasse:
Could you also provide english translations because this is the main language that will be shown if there is no translation in other languages. Also the translation tool only work if we have a english text.
It would also be nice if we replace the text "Login name" or "Anmeldename" with "E-Mail" if the option is activated.

Fasse:
BTW you made the changes in the 4.3 branch. We also need the changes in the master. But I have made huge changes to the project structure so there is something to do.

kainhofer:
Of course, I can provide a proper englisch translation. I already included it in the en.xlm file. Is there anything else where I need to include the Englisch version?

I based my patche on the 4.3 branch, because I have been using it on a live system for two years now. The main branch (origin/master) does not even run here, as it seems to be missing some libraries for bootstrapping. I tried a few weeks ago and again a few days ago, but it simply errored out with missing libraries. Of course that might be due to you shuffling files around, or due to some libraries no longer being available in master.

Regarding the "replacement" of the username with the e-mail: My patch still allows the username, it simply allows the e-mail instead of the username in addition.
The logic is:

- If the given username can be found as username in the database, compare the password for that user [that's where the current behavior ends]
- If the username cannot be found as username, assume it is an e-mail and retrieve all users that have it set as username. If there is a user, check the passwort for that user (the username is retrieved by the check before).

So, yes, it might make sense to change the strings from "Login name" to "Login name / E-Mail", but we should not replace with with only "E-Mail", as the username will always keep working.

Fasse:
I'm sorry, didn't saw that you already provided an English translation, so everything is fine.

If you use the master you must run `composer install` so all libraries will be installed. After that the master version should work.

Fasse:
Thanks for the explanation of the e-mail login usage. In that case I would stay with login name and don't change it.

Fasse:
Could you please provide also a merge request for the master. I don't want to add features to the 4.3 branch if we don't have that feature in the master. A auto merge in the master isn't possible because the codebase is very different.

adm_program/installation/db_scripts/preferences.php
```diff
@@ -16,6 +16,7 @@
     // System
     'enable_rss'                     => '1',
     'enable_auto_login'              => '1',
+    'enable_login_emailaddress'      => '0',
```
Fasse, line 19: for the new preferences we use a prefix with the preference area so `security_login_email_address_enabled` would be a better name

kainhofer:
Thanks for the feedback. I renamed the preference as requested.

Merge Request for master is available as #1718

Fasse (approved):

Fasse:
That looks good. Thank you very much for that enhancement!

PR #1723, opened 2025-01-02
Implement generic changelog functionality: Hook in the log writing directly in the TableAccess base class for all DB records

kainhofer:
A very first preview for first feedback. This PR is not meant to be merged just yet!
Please feel free to try it out (ideally on a new installation) and give feedback about the general approach. There are lots of bugs/issues still open, but the architecture should be right before proceeding with the tedious tasks of polishing all details and testing everything.

I've based my development on the v4.3 branch, because I prefer working with a stable basis for such large undertakings. Porting over later to a new development branch -- even with lots of changes -- is easier in my experience.

### Approach

By default, whenever a DB record is saved, all modified columns are logged. There is an exclusion list for tables that don't need logging (changelog, autologin, sesssion, etc.), as well as a table-specific exlusion list for column (by default all _{usr_id,timestamp}_{create,change} columns are ignored, but other tables like _users have some more).

Storing all changes in the adm_log_changes table is implemented. For most use cases (users, user_data, members, profile_fields, events, announcements, roles) the logged data is also properly polished. For other DB tables, logging works, but each and every column is logged as a change (except _uuid, _{usr_id,timestamp}_{create,change}). For most cases, this also works just fine.

Also extended the profile history page to a full changelog page for all different record types, properly linking to most records. Using filters, that page can also show only changes to one (or more) DB tables and records (e.g. changes for one user or one Role or one event).

Also added a change history button to the user profile, profile field definitions, memberships, roles, announcements and events. This should give a pretty good picture how and where a changelog can work just fine.

The changelog page adm_program/modules/contacts/profile_field_history.php needs a lot more polishing, though.

### Things to note / open questions

- Update scripts from previous versions are not yet implemented, so all former history is lost currently
- Permissions to view changes are not yet properly implemented. I simply left the 'current user is allowed to view profile chagnes' logic.  Also, we might need to extend the preferences, which changes to log (user/membership data, admidio configuration/preferences, announcements, events, etc.).
- Similarly, I have not thought about multiple organizations.
- In the history page profile_field_history.php, I need to create links to the different modules: Is there any generic function to create a link e.g. to the user profile page of user UUID, or to the group member page of group $id, etc.? For now, I had to hardcode the URLs in profile_field_history.php function createLink...
- In the adm_log_chagnes DB table, each change stores the DB column, which is then shown to the user in the history/changelog page -> Is there any function to return a human-readable name for columns like usr_login_name, mem_begin, rol_description, etc? The same goes for DB table names (users, user_data, members, etc.)
- Having the changelog page profile_field_history.php in the contacts module doesn't make sense for a generic approach. Where shall I move it and what is a better and consistent name?
- I had some dependency problems, so as a nasty workaround, I added some calls like the following in some files. How can this be fixed?
      include_once __DIR__ . '/TableLogChanges.php';
- 

As a sidenote, I had to fix the installation.php script, as it never assigned the $gDb and $gProfileFields global variables. So far, the installation didn't end up in code pathes that needed the profile fields in the User object, but the logging needed it.


kainhofer:
Since many files are affected, I should probably give a little overview of the relevant code parts:
- The main logging functionality happens in the TableAccess class in the save() and delet() methods, which call some helper methods for creation, deletion and modification. They create a TableLogChanges object that stores the log entry in the database.
  - Child classes can modify how fields are stored (e.g. the new TabeUserData does not lot each of its columns as a separate change, but as one change for the underlying user).
- The new adm_log_changes table holds the ids of the affected records (and potentially related objects, like the group for the members table), as well as human-readable strings from the time of the change. Reasons are that (1) for things like membeships, the underlying Db record might be deleted meanwhile and (2) to me a changelog should be immutable and show the status at the time of the change. 
- Most changes to the modules are just inserting a "change history" button
- The changes to [ChangeNotification.php](https://github.com/Admidio/admidio/pull/1723/files#diff-b1c40054a70a35853608b0ee28ae463039ddec825300e3985b999babf2540e52) and TableMembers are simply removing the previous logging functionality (but still leave the email-notifications).
- The profile_field_history.php is extended to show changes to all kinds of db tables (filters are passed in as GET variables table, uuid, id and related_id)

demo_data/db.sql
```diff
@@ -41,14 +41,14 @@ DROP TABLE IF EXISTS %PREFIX%_sessions             CASCADE;
 DROP TABLE IF EXISTS %PREFIX%_texts                CASCADE;
 DROP TABLE IF EXISTS %PREFIX%_user_relations       CASCADE;
 DROP TABLE IF EXISTS %PREFIX%_user_relation_types  CASCADE;
-DROP TABLE IF EXISTS %PREFIX%_user_log             CASCADE;
```
Fasse, line 44: We shouldn't delete this line now because we want to simulate the update process from user_log to log_changes

demo_data/data.sql
```diff
@@ -2941,13 +2941,13 @@ INSERT INTO %PREFIX%_user_data (usd_id, usd_usr_id, usd_usf_id, usd_value) VALUE
 -- Data for table adm_user_log
 --
 
-INSERT INTO %PREFIX%_user_log (usl_id, usl_usr_id, usl_usf_id, usl_value_new, usl_value_old, usl_usr_id_create, usl_timestamp_create, usl_comment) VALUES
```
Fasse, line 2944: We shouldn't delete these lines now because we want to simulate the update process from user_log to log_changes

adm_program/system/classes/TableAccess.php
```diff
@@ -188,6 +191,151 @@ public function countAllRecords(): int
...
+    /**
+     * Retrieve the list of database fields that are ignored for the changelog.
+     * Some tables contain columns _usr_id_create, timestamp_create, etc. We do not want
+     * to log changes to these columns. Subclasses can also add further fields 
+     * (e.g. the users table stores and auto-increments the login count, which 
+     * we do not want to log)
+     * 
+     * @return true Returns the list of database columns to be ignored for logging.
+     */
+    public function logIgnoredColumns(): array
```
Fasse, line 232: @In my opinion `getIgnoredLogColumns` is a better name, because it tells better what the function does.

Fasse (changes requested):
In general I have a problem to test your changes because in my installation I have an old 4.3 database and want to update to your branch. I could not update because I got errors that the table log_changes doesn't exists. Classes like the session want to write something in that table because the session data is changes. 
Also I could not insert demo data because of the same problem.

For demo data and maybe in some other cases we need a mode to disalbe log completly for the current script execution.

kainhofer:
> In general I have a problem to test your changes because in my installation I have an old 4.3 database and want to update to your branch. I could not update because I got errors that the table log_changes doesn't exists. 

It seems I didn't understand how demo data was supposed to work, so I messed up the demo data. I now understand that the demo data (including the copy of db.sql) provides a snapshot of an installation with Admidio version 4.2.12. That data is not supposed to be kept in sync with current code changes. Instead, the update code is supposed to run on it... 
Once I had understood this, it was easy to fix the demo data (just revert all changes and add the defijnition of TBL_USER_LOG).

Installing Demo data should work now. I also properly implemented the DB transformation for the new log table, so old log entries should be preserved, too.

demo_data/db.sql
kainhofer: you are right, that was one of the changes I had to revert...

demo_data/data.sql
kainhofer: As I said, I misunderstood the demo data fundamentally... reverted. And the update process should be implemented.

adm_program/system/classes/TableAccess.php
kainhofer: changed.

Fasse:
Thanks for the changes. The build of the demo data now works and also the update is working for me!

Here are a few points I stumbled across:

- English translation strings not available (you must not provide the german strings, but the english string are necessary for translation process)
- if I change e.g. the name of an event, then also the dat_begin and dat_end are logged because the logging things something at the seconds changes. Old value `2025-01-14 17:00:00` new value `2025-01-14 17:00` . The same with the booleans old value is empty new value is `0`.
- a general decision: Should be log the creation of an object. In my opinion it's not necessary because we log the timestamp of the creation in each table and after that we log all changes. So you are able to recover the values of the creation. 
- Also we need a matching from the internal database fields to user friendly field name. Now it's logged that field `#dat_all_day#` is changed but I want to read `Ganztägig`
- btw  in my opinion we need to add preferences to each module if we want to log the changes. Or what do you think about that. Alternative we have one global preferences to enable logging.

In addition to these comments, I think the function and the current implementation are very good and a good extension of Admidio's functionality.

Fasse:
FYI: we do not need the logging for guestbook module because this module will be removed in version 5.0

kainhofer:
This is superceded by PR #1748.

PR #1731, opened 2025-01-17
Various smaller fixes for the v4.3 branch

kainhofer:
Various smaller fixes

* Typos in the German translation
* missing adjustments due to table renaming adm_dates -> adm_events
* Better handling of regexp and delimiter-detection
* During installation make global variables like $gDb, $gProfileFields, etc. available.
* In the links module, weblinkIsNew was lost in previous commits, leading to unknown variable errors.
* Typos in comments
* Copy&Paste error in array key
* Checking whether a vaiable is empty by converting it to a string does not work for array, so we need to special-case this.
* In TableUserRelationType we need an urt_id_inverse, otherwise we cannot retrieve the inverse.
* In User, when no profile fields are passed, make a clone of the global profile fields object


Fasse (commented):
Looks good. Thanks for the fixes!

Fasse (approved):
Looks good. Thanks for the fixes!

PR #1736, opened 2025-01-20
Fix HtmlDataTable instance naming to avoid conflicts when adding multiple table instances on one page

MightyMCoder:
When trying to add multiple HTMLDataTables to a single HTMLPage the following JavaScript error occurs:
`Uncaught SyntaxError: redeclaration of const admidioTable`

To fix this the `id` of each HTMLDataTable to be added to the page should be included in the constant name.


Fasse:
I also add this to the 4.3 branch, so we have the fix also there.

PR #1743, opened 2025-01-28
fix: '-' in Html(Data)Table Ids prevent using modal windows

### LINKED ISSUE #1741: Couldn't delete files in module documents and files anymore
After applying #1736 it is not longer possible to delete files in  _"documents and files"_  (see [Forum Post 35754](https://www.admidio.org/forum/viewtopic.php?p=35754#p35754)).

#### Problem:
The HtmlTable in _"documents and files"_ is created with `'-'` in the ID:
` $page = new ModuleDocumentsFiles('admidio-documents-files', $headline);`

#### Possible fixes:

1. replace all  `'-'` with  `'_'` in all HtmlTable creations.
2.  replace all  `'-'` with  `'_'` in `HtmlDataTable.php` before adding the ID to the JavaScript constant declaration.

***

@Fasse: What is your preferred fix? What do you think?

MightyMCoder:
fixes #1741: _couldn't use modal windows in HtmlDataTables_ 

***

I also checked all HtmlTable initializations so hopefully I found all occurrences (only 2?).
@Fasse: Maybe you can check again?

### DIFF
--- a/adm_program/system/classes/ModuleDocumentsFiles.php
+++ b/adm_program/system/classes/ModuleDocumentsFiles.php
@@ -153,7 +153,7 @@ class ModuleDocumentsFiles extends HtmlPage
         }
 
         // initialize and set the parameter for DataTables
-        $dataTables = new HtmlDataTables($this, 'documents-files-table');
+        $dataTables = new HtmlDataTables($this, 'documents_files_table');
         $dataTables->disableDatatablesColumnsSort(array(1, 6));
         $dataTables->setDatatablesColumnsNotHideResponsive(array(6));
         $dataTables->createJavascript(count($this->data), 6);
--- a/adm_program/system/classes/ModuleDocumentsFiles.php
+++ b/adm_program/system/classes/ModuleDocumentsFiles.php
@@ -153,7 +153,7 @@ class ModuleDocumentsFiles extends HtmlPage
         }
 
         // initialize and set the parameter for DataTables
-        $dataTables = new HtmlDataTables($this, 'documents_files_table');
+        $dataTables = new HtmlDataTables($this, 'adm_documents_files_table');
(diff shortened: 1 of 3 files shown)


adm_program/system/classes/ModuleDocumentsFiles.php
```diff
@@ -153,7 +153,7 @@ public function createContentList()
         }
 
         // initialize and set the parameter for DataTables
-        $dataTables = new HtmlDataTables($this, 'documents-files-table');
+        $dataTables = new HtmlDataTables($this, 'documents_files_table');
```
Fasse, line 156: Could you add the prefix `adm_` to the ID so we have the naming convention of Admidio 5

adm_program/system/classes/ModuleGroupsRoles.php
```diff
@@ -376,7 +376,7 @@ public function createContentPermissionsList()
         }
 
         // initialize and set the parameter for DataTables
-        $dataTables = new HtmlDataTables($this, 'role-permissions-table');
+        $dataTables = new HtmlDataTables($this, 'role_permissions_table');
```
Fasse, line 379: Could you add the prefix `adm_` to the ID so we have the naming convention of Admidio 5

Fasse (changes requested):

adm_program/system/classes/ModuleDocumentsFiles.php
MightyMCoder: added "adm_"-prefix for table ids with new commit [4dcc5d4](https://github.com/Admidio/admidio/pull/1743/commits/4dcc5d4889f385f56f15209978f8b316382bdfc9)

Fasse (approved):

PR #1744, opened 2025-01-28
fix:  datetime filed didnt show up as disabled

### LINKED ISSUE #1742: Disabled Datetime fields not popssible
When trying to implement disabled form fields of type _"datetime"_ these fields didn't get the disabled property set correctly.

#### Problem:
In _"HtmlForm.php"_ the property only gets set as overall attribute, but when adding a _"datetime"_ field the attribute should be set for the individual date and time field separately.

#### Possible Fix:
Reposition the case structure for adding this property behind the check, if it is a _"datetime"_ -Field and differentiate between adding "_normal"_ and _"datetime"_ field in the case structure:

###### Example:
```
        // set field properties
        switch ($optionsAll['property']) {
            case self::FIELD_DISABLED:
                if ($optionsAll['type'] === 'datetime') {
                    $attributes['dateValueAttributes']['disabled'] = 'disabled';
                    $attributes['timeValueAttributes']['disabled'] = 'disabled';
                } else {
                    $attributes['disabled'] = 'disabled';
                }
                break;

        ...    
```

MightyMCoder:
fixes #1742 _Disabled Datetime fields not popssible_

Fix for the `addInput` method in the `HtmlForm` class to ensure that field properties such as `disabled`, `readonly`, `required`, and `hidden` are correctly applied to datetime inputs.

### DIFF
--- a/adm_program/system/classes/HtmlForm.php
+++ b/adm_program/system/classes/HtmlForm.php
@@ -706,27 +706,6 @@ class HtmlForm extends Smarty
                 break;
         }
 
-        // set field properties
-        switch ($optionsAll['property']) {
-            case self::FIELD_DISABLED:
-                $attributes['disabled'] = 'disabled';
-                break;
-
-            case self::FIELD_READONLY:
-                $attributes['readonly'] = 'readonly';
-                break;
-
-            case self::FIELD_REQUIRED:
-                $attributes['required'] = 'required';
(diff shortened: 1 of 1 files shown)


Fasse (approved):
Looks good. Could you also provide a fix for the master.

MightyMCoder:
I will add a separate fix for master today.

PR #1745, opened 2025-01-31
Fix a missed adm_dates -> adm_events rename

kainhofer:


Fasse (changes requested):
There should also be an sql update for the existing installations in the update script of version 5.

kainhofer:
> There should also be an sql update for the existing installations in the update script of version 5.

It's already there in the v4.3 DB update script:
master: https://github.com/Admidio/admidio/blob/1f039284d78227021d2e234c308b6f81b0c2c042/install/db_scripts/update_4_3.xml#L127
v4.3 branch: https://github.com/Admidio/admidio/blob/1f5864039e36ec6b74f7b8f588144e85436011ff/adm_program/installation/db_scripts/update_4_3.xml#L127


But your are right: It seems that v4.3 is also missing this very patch. The following line also need to be changed:
https://github.com/Admidio/admidio/blob/1f5864039e36ec6b74f7b8f588144e85436011ff/adm_program/installation/install_steps/start_installation.php#L140C56-L141C1

Fasse (approved):

Fasse:
You are right. Didn't saw this.

PR #1753, opened 2025-02-09
More work on the changelog

kainhofer:
Most direct DB access is converted to Entity-based DB access, so these changes are logged automatically.

- Preferences converted to use Entity rather than direct DB access -> Changes are logged. The preference name is not translated into human-readable form, as there is no function for this and there are hundreds of preferences, which I don't want to hardcode.
- Folder and album deletions and renamings use Entity-based DB access; REmoved logging of duplicate information (parent folder and file/folder name is already included in the creation log entry!)
  - Instead of one folder object looping through all child folders and act on other folders, I now create a separate Folder object for each subfolder and tell it to delete itself (and its children). That's more in line with object-oriented programming and also removes some duplicate code.
- List_Columns deletion uses Entity-based DB access and is now logged
- Role_dependencies now also use an Entity-based class to store changes in the DB. Unfortunatel, the table has no auto-increment key (Entity assumes one in keyColumnName!), so some parts are a little more complicated than with other tables.

- Event participation status display improved in the changelog page
- Don't remove log entries when a user is deleted

AFAICS, there is no relevant direct DB access any more for changes that should be logged.

The other issues of the changelog are still open, though. I still don't like the massive code duplication to get nice names for the affected DB columns / tables. Both the translation of the column and table names as well as the display of the values have code duplicated from the preferences module or other core modules, where the human-readable texts are "hardcoded" into the forms....

### What is still missing

- [ ] Write Documentation how to implement/customize proper logging for core and third-party modules -> Should be added to the DokuWiki
- [ ] Multi-organization setup -> Creating/deleting organizations is logged, but the changelog page (and the changelog table!) does not distinguish to which org a record belongs!
- [ ] Access permissions taking into account Organization ID
- [ ] "Delete History" button in the changelog page (nice-to-have)
- [ ] On the Changelog page, provide a table selection filter, potentially add more filters, e.g. by name or old/new value
- [ ] Performance: The changelog page currently loads and formats all changelog entries in the requested time frame. For large installations, this might overwhelm the browser and/or network. Maybe the changelog should be switched to dynamic page loading (like the contact list)?

### Open/Known Bugs (in my changelog code or in admidio in general)

- [ ] Folder rights (table role_rights, internal_name) are not tanslated / user-readable
- [ ] Deleting a folder also logs removal of each role permission
- [ ] Moving a folder removes and re-adds all role permissions
- [ ] Category administration: Type is not used / stored in changelog (EVT, ANN, …) -> Changelog shows all categories, even when not relevant (e.g. calendars are shown for role categories)
- [ ] 	A lot of Code Duplication in ChangelogService: Translating tables, column names and values to human-readable labels -> no generic function, translations are scattered all over the codebase, whenever a form control is created...
  - [ ] DB column -> readable string conversion from individual record editing pages 
  - [ ] Hardcode the html links to the different kinds of tables
  - [ ] Hardcoded code to generate Entity-derived object for a given DB tble (e.g. new User for 'adm_users' table)
  - [ ] Function formatValue contains a lot of duplicate code from user fields -> Write one generic function that formats values, including links to various forms of objects?
  - [ ] $userFieldText to convert user field types to readable strings

### DIFF
--- a/adm_program/modules/changelog.php
+++ b/adm_program/modules/changelog.php
@@ -383,9 +383,9 @@ try {
             if ($row['table_name'] == 'list_columns') {
                 // The related item is either a user field or a column name mem_ or usr_ -> in the latter case, convert it to a translatable string and translate
                 if (!empty($relatedName) && (str_starts_with($relatedName, 'mem_') || str_starts_with($relatedName, 'usr_'))) {
-                    $relatedName = $fieldString[$relatedName];
+                    $relatedName = $fieldString[$relatedName]??$relatedName;
                     if (is_array($relatedName)) {
-                        $relatedName = $relatedName['name'];
+                        $relatedName = $relatedName['name']??'-';
                     }
                     if (!empty($relatedName)) {
                         $relatedName = Language::translateIfTranslationStrId($relatedName);
--- a/adm_program/modules/changelog.php
+++ b/adm_program/modules/changelog.php
@@ -380,6 +380,9 @@ try {
             if ($row['table_name'] == 'roles_rights_data') {
                 $relatedTable = 'roles';
             }
+            if ($row['table_name'] == 'roles_dependencies') {
+                $relatedTable = 'roles';
+            }
             if ($row['table_name'] == 'list_columns') {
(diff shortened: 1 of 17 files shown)


Fasse:
Good improvements to the code!

PR #1768, opened 2025-03-10
Tiny fixes for master (mainly in Presenters / UI elements)

kainhofer:
Lots of tiny fixes / improvements for form / UI generation

### DIFF
--- a/src/UI/Presenter/FormPresenter.php
+++ b/src/UI/Presenter/FormPresenter.php
@@ -898,7 +898,7 @@ class FormPresenter
 
             case self::FIELD_HIDDEN:
                 $attributes['hidden'] = 'hidden';
-                $attributes['class'] .= ' invisible';
+                $attributes['class'] = ' invisible';
                 break;
         }
 
--- a/src/UI/Presenter/FormPresenter.php
+++ b/src/UI/Presenter/FormPresenter.php
@@ -237,7 +237,9 @@ class FormPresenter
         $attributes = array();
         $attributes['type'] = $optionsAll['type'];
         $attributes['data-admidio'] = $optionsAll['data-admidio'];
-
+        if (array_key_exists('style', $optionsAll)) {
+            $attributes['style'] = $optionsAll['style'];
+        }
         // disable field
         if ($optionsAll['property'] === self::FIELD_DISABLED) {
             $attributes['disabled'] = 'disabled';
--- a/adm_program/modules/contacts/contacts_data.php
+++ b/adm_program/modules/contacts/contacts_data.php
@@ -251,7 +251,7 @@ try {
         }
 
         $columnValues['1'] = '<a href="' . SecurityUtils::encodeUrl(ADMIDIO_URL . FOLDER_MODULES . '/profile/profile.php', array('user_uuid' => $row['usr_uuid'])) . '">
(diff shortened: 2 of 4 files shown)


src/UI/Presenter/FormPresenter.php
```diff
@@ -232,12 +232,18 @@ public function addButton(string $id, string $text, array $options = array()): v
         $optionsAll = $this->buildOptionsArray(array_replace(array(
             'type' => 'button',
             'id' => $id,
+            'name' => $id,
             'value' => $text
         ), $options));
         $attributes = array();
         $attributes['type'] = $optionsAll['type'];
+        if (array_key_exists('btn-value', $optionsAll)) {
```
Fasse, line 240: Why do you add `btn-value` to the options array? There is already `value` in the array. What should be the difference beween these two values?

Fasse:
@kainhofer  Could you explain why you want to add `btn-value` ?

src/UI/Presenter/FormPresenter.php
kainhofer: The current 'value' is the display text of the button.

Buttons also have a name and value attribute, which would look like:

`<button id="$id" name="$name" value="$btn-value" class="..." type="submit" data-admidio="">$value</button>`

Currently the generated button looks only:
`<button id="$id" name="$id" class="..." type="submit" data-admidio="">$value</button>`

The button will be included in the submitted $_POST array currently as "$id" => ''. With name and value, it would be "$name" => '$btn-value'.

This is usually used to distinguish which button was clicked when multiple buttons are present. Since value was already in use for the display text, I thought using btn-value would be the next-close choice. The $id is used as the key in the elements list of the presenter, so that must be unique, but multiple controls can have the same name.


However, then I realized that Admidio's JS form submission wouldn't submit the pressed button automatically, as it is no longer the button that triggers the submission. So, in the end I ended up not using this anyway (and removing functionality that depended on multiple buttons in the same Presenter form). Shall I remove this (and also the 'name') from the PR or do you think it might be useful in the future?

Fasse: I think if we don't have a useful implementation and support for multiple form buttons than we shouldn't implement these things because users might think we support it.

If we implement it, than the JS script should also be changed to support this. But then I think we should use `value` and add an optional new parameter `label` or `text`. If only `value` is set than the behavior is like now otherwise the new behavior will be used.

kainhofer: In this particular case, I concur, since the submission JS would need significant change, too. (On the other hand, with custom JS, one could make extensive use of further html attributes and elements...)

However, in general I found the control-generating functions in the PagePresenter currently quite limited. It supports only those attributes that are hardcoded, and any extension needs changes in the core code. With JS, one can do a lot of really nice things UI-wise, but the lack of built-in support for e.g. custom data-* attributes or even standard html attributes, at times makes it quite inflexible...

kainhofer:
Updated PR to remove the name and value attributes from input buttons again. (I'm not using it currently in my SSO development, anyway).

src/UI/Presenter/FormPresenter.php
Fasse: You are absolutly right. A more flexible handling of custom attributes and more standard attributes makes sense. But then this should be implemented for all fields. I will create an issue for that so we won't forget this.

PR #1772, opened 2025-03-16
Single-Sign-On (SAML 2.0): First working version of the SAML 2.0 id Provider.

kainhofer:
Implement SAML id Provider functionality for Admidio. This means that  other apps can offer user authentification and authorization based on Admidio's user baser. With this single-sign-on, the users will be redirectory from the other app ("Service Privider") to Admidio, log in there and be given access to the Service Provider. For now, there is no user/admin documentation yes, but that's on my todo list.

I wasn't sure whether this should be placed in the plugins or modules directory, so I opted for adm_plugins/sso/, but the Entities and Presenters are in the usual location, anyway. The configuration can be found in the preferences -> modules -> SSO. There, public keys can be generated, the SAML idP configuration happens and clients can be setup. Of course, the client (Service Provider) also needs to setup SAML SSO.

* Add SSO key management functionality
* Add SAML client (SP) configuration (from metadata)
  * Metadata, SSO and SLO endpoints
  * user ID field, profile fields mapping, roles mapping, role restrictions
* Add preferences for SAML (and later OIDC)
* Implement metadata, SSO and SLO endpoints.
* Implement response and assertion signing and encryption
* If a user successfully logged in, but group permissions prevent access, show a message on the login screen and allow the user to switch accounts.

* Using Nextcloud, Dokuwiki or Wordpress as SP works, roles are properly transferred (for WP there are no free plugins that support roles/group mapping).
* Fix metadata XML downloading (either through JS/Browser or if CORS issues arise through proxy script in admidio tree)


kainhofer:
I have now created some user/admin documentation at https://www.admidio.org/dokuwiki/doku.php?id=en:2.0:single_sign_on

Fasse (changes requested):
The plugin folder is something where external plugins that do not want to provide code to the Admidio core could add there files. For historical reasons the code of the overview page are in the plugin folder. So if you have something for the overview page, this part should be in the plugin folder.

Since I only look at the code I'm not sure if your code in the plugin folder is only for an overview plugin.

You add the preferences to the module part and not to the common part. Maybe the naming common and module are here not optimal because the common part should include things that do not provide direkt content to the users and needed to configure the system. So I think your new enhancement is something like this because it provides an alternative login but no content for an organization.

Fasse:
BTW: Many thanks for implementing this feature in the Admidio core. I think this will be very useful for many installations. Also your provides documentation reads well and is detailed. With that I will later try to use it for the Admidio homepage too.

PR #1774, opened 2025-03-21
Fix the label for the edit right input field on the category edit screen

nanoflow:
Screenshot 2025-03-21 at 08 49 25


nanoflow:
Screenshot 2025-03-21 at 08 55 19

not sure, if the translation is really fitting since it does not contain anything regarding edit rights. But together with the placeholder it should be clear to everyone

PR #1784, opened 2025-04-05
PHP 8.3: Fix  warnings: if a parameter has a default value of null, it must be declared as nullable

kainhofer:
e.g. if a function has signature funcname(string $param = null), php8.3 will throw a warning, because the type of $param is string. If null is also allowed, one must explicitly declare the type as nullable, either via ?string or string|null.

### DIFF
--- a/src/Infrastructure/Entity/Entity.php
+++ b/src/Infrastructure/Entity/Entity.php
@@ -799,7 +799,7 @@ class Entity
                             $queryParams[] = $value;
                         }
                     }
-                    // Ignore the usr_id_create and timestamp_create (and *_change) columns in the change log...
+                    // Ignore the usr_id_create and timestamp_crearte (and *_change) columns in the change log...
                     if (!in_array($key, $this->getIgnoredLogColumns())) {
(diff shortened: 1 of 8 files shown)


src/Infrastructure/Entity/Entity.php
```diff
@@ -799,7 +799,7 @@ public function save(bool $updateFingerPrint = true): bool
                             $queryParams[] = $value;
                         }
                     }
-                    // Ignore the usr_id_create and timestamp_create (and *_change) columns in the change log...
+                    // Ignore the usr_id_create and timestamp_crearte (and *_change) columns in the change log...
```
Fasse, line 802: @kainhofer this line should be reverted

Fasse (changes requested):

src/Infrastructure/Entity/Entity.php
kainhofer: Right. Unfortunately, that typo also made it into the other PR, which is already merged, so I added the revert here.

PR #1785, opened 2025-04-05
SSO SAML: Multiple fixes and improvements

kainhofer:
* Use redirect binding for Single-Log-Out
* Refactor Client and Service classes to a SSO base class with common features/code and minimal SAML&OIDC subclasses


adm_program/languages/en.xml
```diff
@@ -1396,19 +1396,19 @@
...
+  <string name="SYS_SSO_ENCRYPTION_KEY">Key for encryption</string>
   <string name="SYS_SSO_SAML_ROLES">Included groups/roles</string>
   <string name="SYS_SSO_SAML_ROLES_DESC">Select all roles that should be transmitted upon successful login to the Service Provider. As many clients use the transmitted role verbatim, one can map Adimidio's role names to role names desired on the client / Service Provider's side. If the SAML role is left empty, Admidio's role name is used instead.</string>
-  <string name="SYS_SSO_SAML_ATTRIBUTES">Included user attributes</string>
-  <string name="SYS_SSO_SAML_ATTRIBUTES_DESC">Select all user fields that should be transmitted upon successful login to the Service Provider. For each field, the mapping to the SAML attribute name can be given. If the SAML column is left empty, Admidio's field name will be used instead.</string>
-  <string name="SYS_SSO_SAML_ATTRIBUTE">SAML attribute name</string>
-  <string name="SYS_SSO_SAML_ATTRIBUTES_ALLOTHER">Include all remaining fields with Admidio's internal field name</string>
+  <string name="SYS_SSO_ATTRIBUTES">Included user attributes</string>
```
Fasse, line 1408: @kainhofer In general it would be nice if not all translation strings have the prefix SSO or SAML because the strings should also be used in other scenarios. e.g. SYS_INCLUDED_USER_ATTRIBUTES ..

kainhofer: In this particular case, the string came from a SAML-specific text, which I tried to generalize also for the OIDC SSO. On the other hand, the _DESC of the string is SSO-specific, so I didn't want to tear the string and the corresponding description apart name-wise. In other cases I added generic string IDs *SYS_ADD_ALL_FIELDS", "SYS_ADD_ROLE", "SYS_ADD_ALL_ROLES", "SYS_LOGIN_TO", "SYS_ERROR_UUID_MISSING", etc

PR #1791, opened 2025-05-03
Change getDiskSpace in FileSystemUtils.php to return Integers...

RealDekkia:
...instead of Floats

/src/UI/Presenter/PreferencesPresenter.php calls getHumanReadableBytes (from FileSystemUtils.php) in line 2144 with data from the array returned by getDiskSpace.

Since getDiskSpace returns floats and getHumanReadableBytes requires integers this can cause the settings-page to come up blank.

The easiest fix is changing getDiskSpace to output Integers because storing byte-values as float isn't really necessary anyway.


Fasse:
Thanks for the fix!

Issue #1776, opened 2025-03-26
feat: Show duration of executing or executed role


coolusaHD:
**Is your feature request related to a problem? Please describe.**
As in the most associations we honor the execution of roles after a given number of years. Currently I only see the start and end date but not the duration.

**Describe the solution you'd like**
It would be nice to see for how long a person was/is in the listed role. Also it would be nice to generate a report with that information

**Describe alternatives you've considered**
Currently you have to calculate it with other tools

**Additional context**

PR #1800, opened 2025-05-15
feat: Add showing duration of role in profile

coolusaHD:
See #1776

### DIFF
--- a/system/classes/CategoryReport.php
+++ b/system/classes/CategoryReport.php
@@ -201,6 +201,9 @@ class CategoryReport
                 case 'a':                    //a=additional
                     $number_col[$key + 1] = '';
                     break;
+                case 'd':                    //d=duration
+                    $number_col[$key + 1] = '';
+                    break;
             }
         }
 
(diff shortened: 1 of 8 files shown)


system/classes/CategoryReport.php
```diff
@@ -374,6 +403,12 @@ private function generate_headerSelection()
         $this->headerSelection[$i]['cat_name'] = $gL10n->get('SYS_ADDITIONAL_COLUMNS');
         $this->headerSelection[$i]['data'] = $gL10n->get('SYS_ROLE_MEMBERSHIPS');
         $i++;
+        
+        //Zusatzspalte für die Dauer der Mitgliedschaft
```
Fasse, line 407: Please use english language for description

Fasse (changes requested):
If you like you could also add the membership duration to the list konfiguration. There is a block with membership begin and membership end. Within this block it makes sense to add the "membership duration as an option.

system/classes/CategoryReport.php
```diff
@@ -271,6 +275,26 @@ public function generate_listData()
...
+                    
+                    foreach ($memberShips as $rol_id) {
+                        $role = new Role($gDb);
+                        $role->readDataById($rol_id);
+                        
+                        // Get membership data for this role
+                        $membershipData = $membership->readDataByColumns(array('mem_rol_id' => $rol_id, 'mem_usr_id' => $member));
+                        
+                        if ($membershipData) {
+                            $startDate = $membership->getValue('mem_begin', 'Y-m-d');
```
Fasse, line 291: You don't need these two lines with the dates anymore.

Fasse (changes requested):

Fasse:
@coolusaHD Could you please have a look at the integration in the list configuration. There I only get one number as output if I assign the column "duration of membership". I expect a format like in the profile.

coolusaHD:
@Fasse 
See #1808

PR #1804, opened 2025-05-21
Profile view fixes

MightyMCoder:
- [x] Feature: implement a preference to switch between old and new mode for profile fields visualization
- [x] Bugfix: make relation tab only visible if the relations preference is enabled

see: https://github.com/Admidio/admidio/pull/1789#issuecomment-2895072721


modules/profile/profile.php
```diff
@@ -297,6 +297,11 @@ function formSubmitEvent(rolesAreaId = "") {
                     }
                 }
             } else {
+                // if profile_show_empty_fields is set to 0 then skip empty additional fields
+                if (!$gSettingsManager->getBool('profile_show_empty_fields') && $user->getValue($field->getValue('usf_name_intern'), 'html') === '') {
```
kainhofer, line 301: This check here applies only to profile fields that are not basic data. This means that things like birth date, website, mobile, phone, etc. are still shown, even if empty (because they are not relevant in a professional association like ours - we even don't want too much user data to comply with the GDPR minimal data requirement). That check should be move up, proably before the check for basic data.

MightyMCoder: I thought the basic data should be displayed every time but it's no problem to move this up to also filter BASIC_DATA fields.

PR #1813, opened 2025-06-01
Feature contacts and groups-roles-lists redesign

MightyMCoder:
Implements a select box that lets you filter group members or contacts by status:
- active
- inactive/former
- active and inactive/former

@Fasse: Can you check if the changes made to `ListConfiguration.php` causes any unwanted side-effects?  I haven’t found any myself.

### DIFF
--- a/src/Roles/Entity/ListConfiguration.php
+++ b/src/Roles/Entity/ListConfiguration.php
@@ -240,7 +240,8 @@ class ListConfiguration extends Entity
                     $content = '<span class="' . $buttonClass . '">' . $htmlText . '</span>';
                 } else {                    $content = $htmlText;
                 }
-            }        } elseif ($column->getValue('lsc_special_field') === 'mem_duration') {
+            }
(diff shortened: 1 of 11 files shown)


src/Roles/Entity/ListConfiguration.php
```diff
@@ -899,7 +924,39 @@ public function getSQL(array $options = array()): string
         }
 
         // Set SQL-Statement
-        if ($optionsAll['showAllMembersDatabase']) {
+        if ($optionsAll['showAllMembersDatabase'] && $optionsAll['showFormerMembers']) {
```
Fasse, line 927: Why do we need this part? If I set showAllMembersDatabase then I must be an Administrator and this should also includes all former members.

MightyMCoder: Yes, that's right. This case is no longer needed. 

Before my changes there was only the still existing case with showAllMembersDatabase. This case here was needed to select only former members from other organisations. I forgot to delete this case when I removed the selection of former members of current and other organizations.

PR #1820, opened 2025-06-12
Feat clamp lines

### LINKED ISSUE #1264: Description of Calendar entries to long
**Is your feature request related to a problem? Please describe.**
The description of calendar entries takes to much screen space if the text is very long. 

**Describe the solution you'd like**
The description of calendar entries should be cut at a certain length (like 10 lines) and only be displayed completely after a "extend" button was clicked. 

**Describe alternatives you've considered**
A translate this text button would also help (see below) but it would wirk around the issue instead of to solve it. 

**Additional context**
We translate the description from german to english and add it to the description field after the original language. 
That means our descriptions get very long.

MightyMCoder:
Implement feature to limit the number of visible lines in text blocks (clamping) in event descriptions and announcements.

Closes #1264

### DIFF
--- a/install/db_scripts/preferences.php
+++ b/install/db_scripts/preferences.php
@@ -123,7 +123,7 @@ $defaultOrgPreferences = array(
     // Announcements
     'announcements_module_enabled' => '1',
     'announcements_per_page' => '10',
-    'announcements_clamp_text_lines' => '5',
+    'announcements_clamp_text_lines' => '0',
 
     // Category-Report
     'category_report_module_enabled' => '1',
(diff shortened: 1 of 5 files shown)


Fasse:
I think the number of lines is not a preference of the design panel because this should only have real layout preferences that affects all pages. Why not use within the module a preferences with the lines. Default is 0 and than all is shown. If set to a number than only these lines are shown. Than it's also possible to set module individual number of lines.

The UI of the open and close does not meet my expectations. I would expect a text, something like "more". Most important, the UI to open and close should be at the place where the text stops and not in front of the line where the text stops. The current UI could simply be overlooked if working with Admidio.

MightyMCoder:
No problem. I will create individual preferences for that.

MightyMCoder:
@Fasse:
I redesigned the clamping feature and hope this fits your requirements.

PR #1827, opened 2025-06-18
Feat dropdown table

### LINKED ISSUE #842: Manage dropdown entries
https://www.admidio.org/forum/viewtopic.php?f=2&t=7959

Make a own db table for dropdown entries to mark some entries as old (hide them)

MightyMCoder:
* Added a new db table with columns for option values, sequence, and obsolete status to store selectable options for user fields of type dropdown and radio button.
* Removed the `usf_value_list` column from the user fields table as its functionality is now handled by the new table. 
* Added a new preference `profile_show_obsolete_select_field_options` to control the visibility of obsolete select field options in the profile field edit form.
* Update changelog to track changes in the new table.

#### Advantages:
* Moving options in the profile field edit form no longer affects the selected entries in user data since the options are no longer tied to their position in the former `usf_value_list` column.
* Previously obsolete marked options can now be restored provided the `profile_show_obsolete_select_field_options` setting is enabled.

Closes #842

### DIFF
--- a/modules/profile-fields.php
+++ b/modules/profile-fields.php
@@ -31,7 +31,7 @@ try {
     require(__DIR__ . '/../system/login_valid.php');
 
     // Initialize and check the parameters
-    $getMode = admFuncVariableIsValid($_GET, 'mode', 'string', array('defaultValue' => 'list', 'validValues' => array('list', 'edit', 'save', 'delete', 'delete_option_entry', 'sequence')));
+    $getMode = admFuncVariableIsValid($_GET, 'mode', 'string', array('defaultValue' => 'list', 'validValues' => array('list', 'edit', 'save', 'delete', 'check_option_entry_status', 'delete_option_entry', 'sequence')));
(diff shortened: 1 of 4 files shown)


Fasse:
I like the feature and also your implementation. Only thing I think about is, that's now not possible to delete an entry completely from the database. If I add an entry and than think that it was wrong, I have no chance to delete it.

But I understand your thing, that already existing entries should not be deleted. Maybe we implement an intelligent delete button. I someone click on delete we check if that value for that field is in the database. If not we delete it completely if we found at least 1 value we set it to obsolete. What do you think?

Fasse:
Das funktioniert jetzt soweit. Kannst du aber noch vor dem entgültigen Löschen eine Abfrage einbauen. Das haben wir bisher immer so gemacht, wenn man was löschen will. Falls jemand mal versehentlicht auf den Button drückt, sollte nicht direkt der Eintrag weg sein. Musst du aber nur da machen, wo man jetzt auch sofort löscht. Kannst aber aber auch in beiden Fällen machen, wenn es technisch zu kompliziert ist.

PR #1828, opened 2025-06-18
Feat uniform move actions

MightyMCoder:
Implement uniform move action button behaviour and style to weblinks, CategoriesPresenter, MenuPresenter and ProfileFieldsPresenter.


Fasse:
Unify the handling of the move action is a good idea. Wouldn't it be useful also unify the UpdateXYMoves() function?
You can give this function all necessary information via parameter and implement that function in our basic JS file.

PR #1829, opened 2025-06-28
Inventory module

### LINKED ISSUE #35: Inventory management system
Tracking Bug
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10377331-inventory-management-system?utm_campaign=plugin&utm_content=tracker%2F10474012&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10474012&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1047: Option to show all members of a list
Now it's only possible to show the lists within the following number of entries at once 10, 25, 50 or 100.

We should also add an option to show **all** entries. 

This is possible through a parameter setting:
https://datatables.net/reference/option/lengthMenu

MightyMCoder:
Closes #35 
Closes #1047


MightyMCoder:
@Fasse: If you have no objections i will merge this PR today.

Fasse:
I want to have a deeper look into this but the weather is to hot for computer time :)

One thing I found: I the module is deactivated than it should not be shown within the menu. Now it's shown and I got a message if I click on it.

MightyMCoder:
> I don't understand why I have the checkbox "inventory" when I borrow an object. In my case the object is in inventory and I want to borrow it to another person. So I think I set the person and the borrow date. But per default I must fill the return date. To avoid this, I have to uncheck the "inventory" checkbox and then the borrow date is a must. 
> 
> 
> 
> But why so complicated?


The checkbox indicates whether the item is borrowed or not. When it is not borrowed ("in inventory" is checked) I have to fill out a "received back" date if the item has a "last receiver" and a "lend on" date is set.

When I lend an item (uncheck "in inventory") an nothing else is set the "last receiver" and the "lend on" date are required. 

The logic behind that is that when an item was borrowed to anyone and I mark it again as back in inventory there has to be a received back date set.

Fasse:
I have merged this. We can discuss about some functions later.

PR #1831, opened 2025-07-08
Feat bulk edit delete contacts

### LINKED ISSUE #1285: Mass edit members
Hi,

I´ve imported members via Excel and it imported 500 empty members and I can´t find any mass delete field by selection all 25 or 100 entries per page in order to delete them. Such a feature would be very nice. 

Is it easy to delete these users by phpMyAdmin or has every user entries on different tables?

MightyMCoder:
Closes #1285 

- Also final implementation for inventory module.
- fix: sidebar not updated after saving module visability -> Now there is a redirect to the saved panel to force a page refresh.


Fasse:
How does the bulk edit / delete work? Can't find the functionality.

MightyMCoder:
image

MightyMCoder:
> Ah ok. But why I couldn't edit or delete when all active members are selected?

It would be possible to add the buttons and the select boxes also to all other select filters if this is preferred.

MightyMCoder:
Do you think it make sence to also show the buttons and the checkboxes when filtered for all organizations?

themes/simple/templates/sys-template-parts/form.input.tpl
```diff
@@ -12,14 +12,24 @@
...
+                        {$data.label}
+                        <input class="form-check-input focus-ring" type="checkbox" id="toggle_{$data.id}" name="toggle_{$data.id}" data-bs-toggle="tooltip" title="{$l10n->get('SYS_FORM_USE_FOR_EDIT')}" value="1"tabindex="-1" autofocus="false">
+                    </label>
+                </div>
+            {else}
+                <label for="{$data.id}" class="{if $formType neq "vertical" and $formType neq "navbar"}col-sm-3 col-form-label{else}form-label{/if}">
+                    {include file="sys-template-parts/parts/form.part.icon.tpl"}
+                    {$data.label}
+                </label>
+            {/if}
```
Fasse, line 29: If this is exactly the same code in every 5 form elements than we can extract this to a separate tpl in the parts folder.

Fasse (changes requested):
Great work @MightyMCoder . Some small notes from me in the code part.

Issue #1832, opened 2025-07-11
DataTables auto type detection


MightyMCoder:
I noticed some weird behavior of the dataTables when redrawing the table I think we have to solve somehow.
When serverside processing is not set the dataTable automatically determinates a auto-type for the column when initializing. So when we define a column sort e.g. left we set the className for this column to `text-left`. But DataTables automatically detects the data as numeric and add the class `dt-type-numeric` to it what is basically `text-right`:

<img width="826" height="910" alt="image" src="https://github.com/user-attachments/assets/e3b736a2-c7f6-47c3-a68f-97b1c4f53d4f" />

If serverside processing is enabled then this doesn't happen directly because then DataTables autodetects the types before the processing is finished resulting in no additional classes added. But when then redrawing the table with JS or by ordering the column the autodetection hits in and add the classes sometimes wrong. You can reproduce this in contacts module by clicking the select All checkbox or just order by any column. Then the columns `No.`, `Username` and `Gender` gets autodetected class `dt-type-numeric` set.

Before ordering:
<img width="1400" height="506" alt="image" src="https://github.com/user-attachments/assets/2a7457f5-b43c-4a4c-abc8-ad228e8b27fe" />

After ordering by any column:
<img width="1409" height="483" alt="image" src="https://github.com/user-attachments/assets/f2ddac0f-d1b9-4e50-bbac-e78edb16aa2c" />

Maybe we should consider eighter prevent the auto type detection in general or use it the right way. There are possibilities to use these types to enhance the ordering for example. Currently the ordering is based on plain strings in most cases.

MightyMCoder:
@Fasse:
I figured out where the problem came from.
We often use left/right as column alignment values. These are necessary for the style attribute `text-align`. The DataTables uses their own css classes called` text-start`/`text-end`.
By simply replacing left/right with start/end in the DataTables `setColumnAlignByArray` method this Problem is fixed.

Maybe there is a better solution for this but as long as we have some modules using both DataTables and pdf exports we need to use such a workaround in my opinion.

PR #1845, opened 2025-09-05
Remove deprecated classes

MightyMCoder:
- replace deprecated class `HtmlTable` with smarty template solution
- replace deprecated class `HtmlPage` with `PagePresenter` class
- small improvements

Closes: #1832


Fasse:
We should not remove the deprecated classes. They are set deprecated in this version, so we should deliver these classes and remove them in 5.1 or 5.2 . Maybe some plugins still use these classes.

Fasse (approved):

PR #1857, opened 2025-10-06
Use translation-ids for user relation types

Fasse:
When preparing new demo data I thought it would be useful if we support translation-IDs for user relation types. Than we can deliver this for all installations and the user relation types will always be in the language of the user.


MightyMCoder (approved):

MightyMCoder:
I haven't found any issues so I think this is ready to merge.

Issue #1867, opened 2025-10-12
Category Report: "Duration of Memberships" column does not list roles with multiple memberships (past and present membership)


kainhofer:
**Describe the bug**
In the category report the "Duration of membership" column should list the duration of ALL current memberships. If a user was member of a role some time in the past, and then became a member of that role for a second time, that current role will not be included in the durations.

**To Reproduce**
1. set up a category report config with the "Duration of membership" (internal ID "ddummy") content
2. Add/modify a user so they had a role assigned in the past with an end date, and a second role membership of the same role, which is still valid
3. Create the corresponding category report
5. See error: The "Role: ..." column type (rNN type) will list the user as a member, the "Role memberships" (a type) will show the user as a member, but the "Duration of membership" (ddummy type) column will not list that role membership.
**Expected behavior**
The role membership of the role with a past and a present membership should be included.

**Screenshots**

<img width="494" height="848" alt="Image" src="https://github.com/user-attachments/assets/252f40f8-0364-4c00-9c09-9bf081b5545b" />

<img width="497" height="338" alt="Image" src="https://github.com/user-attachments/assets/7af276e3-974b-4aa2-8e83-8342cddd5722" />

**System (please complete the following information):**
 - Admidio-Version: git master

**Cause of the problem**
Cause of the problem is that the category report uses direct sql statements to check role memberships (so the rNN and a types are correct), but to extract membership data, it calls  
`$membership->readDataByColumns(array('mem_rol_id' => $rol_id, 'mem_usr_id' => $member));`
 which is documented to return false if two or more DB entries with the role and user id are found.

Unfortunately, I could not find any method in the User, Membership or Entity classes that could deal with double entries or is based on the mem_id as opposed to the (non-unique) combination or rol_id and usr_id.

PR #1868, opened 2025-10-12
CatRep: UI improvement; add columns for former roles, membership start/end/duration, user profile fields (uuid, login_name, last_login, etc.)

kainhofer:
* Add new column type f111 to list former members of the role. Although the CatRep by default only lists current membes, those former members will be included, but the former members column will be marked correspondingly to prevent confusion.
* Add new column types b111, e111, d111, uuuid, ulogin_name, ulast_login, unumber_logins
* Fix bug where category restrictions in the category report would not be applied at all
* Split catrep configuration column type select box into two boxes, where the first allows user to select the role, and the second allows the selection of the role membership property (current/past member, leader, start, end, duration).

NOTE: The bNN, eNN and dNN column types still suffer from the multi-membership issue described in issue #1867. The call to $membership->readDataByColumns returns false if a user has two or more memberships (past and present) on the same role.

image
image


Fasse:
I don't know if it's from your changes but anyway we should only work with the group roles in this report and not with the event roles. Because of this the membership column should only show group roles:
Bildschirmfoto 2025-10-20 um 12 24 07

kainhofer:
I didnt't change which roles are considered in either the role selection box, nor the displayed memberships with the adummy or ddummy column types.

Do you mean to exclude all event roles from the category report at all? I.e. shall the event roles be removed from the list of column types (the select box in the configuration), too? Or shall they only be excluded from the adummy and ddummy column types?

Fasse (approved):

Fasse (changes requested):

PR #1890, opened 2025-11-14
Resolve issues

RemyLoveLogicAI:
Please fix the following security issue:

<issue>
Information disclosure via phpinfo function output exposed in PHP application

Attackers can read server configuration, versions, environment variables, and file paths, enabling targeted exploitation, reconnaissance, and privilege escalation.

The code calls phpinfo() in a reachable endpoint, exposing verbose environment and configuration details to unauthenticated users.
</issue>

<locations>
/tmp/sourcery/RemyLoveLogicAI/admidio/48f58d1c-0c85-4eb9-93bf-4eca852bcb35/adm_program/system/phpinfo.php:23
</locations>

<fix>
Remove phpinfo() from production code. If diagnostics are needed, use php -i on the command line or restrict a debug endpoint behind strong authentication and IP allowlisting. Ensure it is disabled in production environments.
</fix>

<fix_impact>
Debug pages that relied on phpinfo output will no longer be available in production.
</fix_impact>

Keep the changes minimal - only make the necessary code changes to fix the security issue.


Fasse:
The mentioned file is not accessible until you have a valid login and the administrator right. So there is no security issue.

PR #1897, opened 2025-11-14
Fix various issues with DataTables (changelog with NULL values, language unset after installation, modal popup boxes) (fixes #1893 and #1895)

### LINKED ISSUE #1893: Changelog not working in contacts
If the Changelog is enabled for profile data but on of old or new value is NULL, no Changelog data gets displayed. 

An error is thrown when trying to retrieve the html value in `changelog_data.php` but this error didn't show up and the table leaves empty (type mismatch: int|string assumed, NULL given). 

@Fasse:
We need to make sure that NULL can be handled in the getHtmlValue method of ProfileFields valueObjects class and in the Changelog the error should be displayed.

### LINKED ISSUE #1895: Contacts list on new 5.0 installation gives datatables error (trying to load datatables..json, since language appears empty)
**Describe the bug**
On a fresh installation from the v5.0 branch, going to the contacts list gives a Javascript error:

Image

The corresponding JS code on the page is:
`const admidioTable_adm_contacts_table = $("#adm_contacts_table").DataTable({"language": {"url": "https://admidio.local/libs/datatables/language/datatables..json", "emptyTable": "No entries available."}, ...`
`

Notice the "datatables..json" with two periods. The problem appears in src/UI/Component/DataTables.php, lines 132 and 134, when no language code is set yet:

```        
if (!empty($this->emptyTableMessage)) {
  $this->datatablesInitParameters[] = '"language": {"url": "' . ADMIDIO_URL . FOLDER_LIBS . '/datatables/language/datatables.' . $gL10n->getLanguageIsoCode() . '.json", "emptyTable": "' . $this->emptyTableMessage . '"}';
} else {
  $this->datatablesInitParameters[] = '"language": {"url": "' . ADMIDIO_URL . FOLDER_LIBS . '/datatables/language/datatables.' . $gL10n->getLanguageIsoCode() . '.json"}';
}
```

Resetting the language code to empty is not possible, once a language has been set, so this is only a problem immediately after installation, until a language is set for the first tim.

kainhofer:
1) There are cases when changelog entries of profile fields can be null, so we need to handle this case explicitly in the function signature and format it as an empty string in HTML for the changelog.

2) Immediately after a fresh installation, no language is set. DataTable, however, tried to load the translated .json by simply inserting the iso code into the url, leading to a broken URL.

As a simple fix, default to 'en' if no language is set yet. Once a language is set, it cannot be unset manually in the settings.

Using the inline operator ?: would be shorter, but using a separate if clause makes the check more explicit and easier to read in the long term.

3) Error handling of DataTables in JS was sub-optimal (expected different structure than handleException, showed a modal popup box, etc.). Switched the error handling to showing the exceptions in the <div id="DT_notice"> block.

Fixes #1895
Fixes #1893

### DIFF
--- a/modules/changelog/changelog_data.php
+++ b/modules/changelog/changelog_data.php
@@ -469,5 +469,8 @@ try {
 
     echo json_encode($jsonArray);
 } catch (Throwable $e) {
-    handleException($e, true);
+    // NOTE: DataTables expects the form {'error' => 'message'}, so we can't use the default handleException($e, true); call!
+    $jsonArray['error'] = $e->getMessage();
+    echo json_encode($jsonArray);
+    exit();
 }
--- a/modules/contacts/contacts.php
+++ b/modules/contacts/contacts.php
@@ -348,6 +348,7 @@ try {
 
     $contactsTable->createJavascript(0, count($columnHeading));
     $page->assignSmartyVariable('headers', $columnHeading);
+    $page->addHtml('<div class="alert alert-danger form-alert" id="DT_notice" style="display: none;"></div>');
     $page->addHtmlByTemplate('modules/contacts.list.tpl');
     $page->show();
 } catch (Throwable $e) {
--- a/modules/contacts/contacts_data.php
+++ b/modules/contacts/contacts_data.php
@@ -477,6 +477,7 @@ try {
 
     echo json_encode($jsonArray);
 } catch (Throwable $e) {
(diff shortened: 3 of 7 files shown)


kainhofer:
I have now found the proper solution. The error is displayed in the `<div id="DT_notice">`, the null values are properly handled, and unset languanges are also handled.

This fixes all three DataTable uses (changelog, contacts and messages).

PR #1961, opened 2026-02-04
Add new language tamil

Arun-TAK:
This PR introduces the new plugin with Billing, Payment, Charges, and REST API modules, along with Tamil language support for the site.

Could you please review the new plugin and proceed with adding it to the official plugins list, including Tamil language support?

Please let me know if any further changes are required from my side.


Fasse:
Thanks for sharing this great plugin. But we will integrate in the Admidio core source code only plugins for the overview page and only those that are maintained from the Admidio core team. 

So please feel free to create your own repository for that plugin like this is done for other plugins e.g. https://github.com/nanoflow/search_member or https://github.com/rmbinder/FormFiller . After that we can give you access to our wiki page https://www.admidio.org/dokuwiki/doku.php?id=en:plugins:index to link your plugin there and to announce that plugin.

After that we should look which components of your plugin should be integrated into the Admidio core e.g. your translation. also interesting your implementation of the API. this could also be a part that we should integrate into the Admidio core.

Arun-TAK:
We’ve separated the resident plugin code into its own repository. This pull request now contains only Tamil translation changes and is ready for merge.

Here is the resident plugin repository: https://github.com/adhi-software/residents

Fasse:
Have you created a new page for your plugin?

Arun-TAK:
How do we create a new page ?

PR #2005, opened 2026-03-24
New vercel 

Sandydiv3r:


Fasse:
I would reject this PR because Vercel is a specific hosting platform rather than a general requirement for Admidio. Adding Vercel-specific support would introduce platform-dependent code and maintenance overhead without providing a significant benefit for the majority of Admidio installations.

Admidio is designed to run on standard PHP hosting environments, so I would prefer to keep the codebase platform-agnostic and support generic PHP/web server environments instead of adding provider-specific integrations.

PR #2010, opened 2026-04-01
feat(contacts): add support for suborganization member sharing

coolusaHD:
### Reason for this PR
The usage of admidio in my town got around and now another organisations wants the same. So I want to try to host them both on the same instance via the sub organisational feature. But I realized that even when the new sub orga is set not to be a suborga anymore I still recieve the hint when adding new member that they are existant in the other non-parent tenant.

This PR fixes that problem and adds a new setting which can be activated in the organisation settings page

### New behavior
With the setting deactivated or the newly no-sub-anymore organisation set up, adding the same mebers dont show any hints anymore.

@Fasse I would love to get your opinon on that

Thanks in advance

### Change Description
- Introduced a new setting to allow suborganizations to use the same members as their parent organization.
- Updated the contacts selection logic to include an option for showing all contacts based on the new setting.
- Enhanced the organization edit form to include a checkbox for enabling/disabling the suborganization member sharing feature.
- Modified various services and presenters to respect the new setting when searching for similar users and managing contacts.


Fasse:
Ok, I understand that feature.

But I think, that organizations that have no relationship should per default not view members of other organizations. If you have a relationship than the parent organization could set this new settings and the childs will inherit that but could not set this for themselves.  

After that we need a new method in `Admidio\Organizations\Entity\Organization` similar to `getFamilySQL` to get all organizations that will share users.  After that the implementation of getting the correct similar users should be done in `searchSimilarUsers` so there is no need to implement it twice.

coolusaHD:
@Fasse I refactored my changes and took your comments into consideration

Fasse:
Sorry for my missing response. I was busy the last weeks. Hopefully I can have a look at it this week or next week.

coolusaHD:
@Fasse Friendly reminder

Fasse:
It looks good to me. I will merge it and do some testing. If there are issues, I will create new ones and assign them to you. 

Thanks for your implementation.

PR #2042, opened 2026-05-04
Life Membership calculation and Years of Membership

Rvice:
Based on selected roles for Years of Membership and Years Towards Life Membership with a configurable number of years offers a means for the category report to let us know when to upgrade a member to Life Membership.

Not all organizations have this feature, but I do know of a few social clubs that do (yacht club and dive club for instance) so I can see this being valuable.

It could get built out even further into a notification for the officers to prepare life member certificates/plaques and things. But for now, this just helps in giving a report that we could then look at in Excel when finalizing the year.

Forum threads:
1. [Life Member frist](https://www.admidio.org/forum/viewtopic.php?p=36877)
1. [Life Member second](https://www.admidio.org/forum/viewtopic.php?p=36894)
1. [Years in Role](https://www.admidio.org/forum/viewtopic.php?p=35437)


Fasse:
Thank you for your contribution and the effort you put into this pull request.

After reviewing the proposed changes, I don't think this feature is a good fit for the general purpose of the Category Report. It introduces a very specific use case that is not aligned with the intended scope of this module. Adding more specialized functionality like this would gradually make the Category Report more complex and harder to maintain.

A better approach would be to implement this functionality as a plugin. This would allow organizations that need this specific behavior to use it without increasing the complexity of the core application.

In addition, starting with Admidio 5.1, the Category Report already provides a membership duration column, and conditions can be applied to report columns. This should cover many scenarios that previously required custom extensions while keeping the report flexible and generic.

For these reasons, I don't think this feature should be merged into the core application.

Issue #2079, opened 2026-07-19
SSO / OIDC URLs miss "oidc/" resulting in {"error":"URL or authorization protocol not available"}


Backtrennpapier:
**Describe the bug**
Using the SSO module with OIDC, on the other side using Nextclouds user_oidc app. 
(https://www.admidio.org/dokuwiki/doku.php?id=en:2.0:single_sign_on:oidc_nextcloud)

First I thought it was just a minor interface issue while at this: the interface shows the incomplete Discovery URL missing the "oidc/". In the documentation the correct URL is cited https://[YOUR_ADMIDIO]/modules/sso/index.php/oidc/.well-known/openid-configuration

Though the issue seems to go deeper. When I corrected the user_oidc app configuration with the correct discovery URL I could save it.

But when trying to login into nextcloud using Admidio the wrong URL (https://sub.domain.de/modules/sso/index.php/.well-known/openid-configuration) is used (missing the /oidc) and the error "{"error":"URL or authorization protocol not available"}" is shown. 

**Expected behavior**
- The discovery endpoint should have the corrected URL https://sub.domain.de/modules/sso/index.php/oidc/.well-known/openid-configuration
- The other endpoints should have the corrected URL https://sub.domain.de/modules/sso/index.php/oidc/userinfo

**System (please complete the following information):**
- Admidio-Version 5.0.11
- PHP 8.4.23
- Database MariaDB

**Might relate to**
- https://github.com/Admidio/admidio/issues/1901#issue-3630302756
- https://github.com/Admidio/admidio/issues/2078#issue-4922694395

Backtrennpapier:
I did two changes which solved I for me, so I could successfully configure the Client and Provider; first the client call to the provider had the same issue as stated in the title, but now it works seamless:
- I tried the modification suggested in https://github.com/Admidio/admidio/issues/1901#issue-3630302756 , but this did not work
- I switched to PHP 8.5(.8)
- I modified the URLs in the database manually, see https://github.com/Admidio/admidio/issues/2078#issue-4922694395
Now it works without issues.

PR #2101, opened 2026-08-03
SSO OIDC: Fix default issuer and endpoint URLs

### LINKED ISSUE #2078: SSO / Changing base-url does not change OIDC URLs
**Describe the bug**
I recently needed to change the base URL for the administration, which can be done easily via the organization settings. But it seems that changing the base URL does not update the OpenID Connect URLs. The SAML URLs are updated.
Looking into the database, table add_preferences the sso% prf_names seems not the updated. Funny enough, it is still old.domain.de for both sso_saml_entity_id and ssh_oidc_issuer_url (although the SAML metadata URL is updated in the frontend).

**To Reproduce**
1. Go to Organisation settings
2. Change old URL to new URL e.g. old.domain.de to new.domain.de
3. Save, Logout, Refresh, verify that new URL is used
4. Login, Go to SSO settings

**Expected behavior**
When updating base URL, the preferences for the SSO provider might require an update too.

**Screenshots**

**System (please complete the following information):**
 - Admidio-Version 5.0.11
 - PHP 8.4.23
 - Database MariaDB

Image

kainhofer:
Derive the default OIDC issuer from the complete SSO route instead of falling back to the Admidio base URL. This restores the missing /oidc path in discovery, authorization, token, UserInfo and JWKS URLs.

When the default value is not changed, instead of storing the default value in the database, store an empty string and let the OIDCService class interpret empty issuer URLs as default value. When an installation is moved to a different (sub)domain, the admidio OIDC-part will adjust accordingly.

In the OIDC settings, display the resolved issuer in the SSO settings as a placeholder text if the input field is empty, but avoid persisting the default so that later changes to the Admidio base URL are reflected automatically.

Preserve explicitly configured issuer URLs and normalize trailing slashes. This is only an issue for OIDC (where actual URLs are configured), and must NOT be applied for SAML (where the entity ID is really a unique identifier and not a URL).

To show the placeholder text and also copy it with the copy icon next to the input field, the JS handler is extended accordingly. This functionality can be used in any input field by simply adding the 'placeholder' property when the 'copy-container' css class is assigned.

Fixes #2078 and #2079

### DIFF
--- a/languages/de.xml
+++ b/languages/de.xml
@@ -1607,7 +1607,7 @@
   <string name="SYS_SSO_OIDC_ENABLED">Single-Sign-On über OpenId Connect (OIDC) aktivieren</string>
   <string name="SYS_SSO_OIDC_ENABLED_DESC">Diese Einstellung macht Admidio zu einem OIDC-Identitätsanbieter (idP) für Single-Sign-On, der es anderen Webanwendungen ermöglicht, Benutzer:innen gegenüber der Admidio-Benutzerliste zu authentifizieren.</string>
   <string name="SYS_SSO_OIDC_ISSUER_URL">Aussteller</string>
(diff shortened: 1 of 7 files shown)


Fasse:
Does this also fix a security advisory of the SSO module?

kainhofer:
No, unfortunately not. This patch is only about the settings and which values are stored in the DB and which are derived from the current host name. This patch does not change any of the security-related code parts.

PR #2121, opened 2026-08-21
Installation: Create the forum tables on MySQL and MariaDB

### LINKED ISSUE #2120: The forum DB tables are never created on a fresh MySQL or MariaDB installation
`install/db_scripts/db.sql:269` and `:290` end the two forum table definitions with

```sql
    ENGINE = InnoDB
    DEFAULT CHARSET = utf8mb4
    ENCODING 'UTF8';
```

`ENCODING` is PostgreSQL syntax for `CREATE DATABASE`; MySQL and MariaDB reject it. Every other
table in the file ends with `COLLATE = utf8mb4_unicode_ci` instead, and these are the only two
occurrences of `ENCODING` in the whole file.

Verified against the MariaDB used by the test suite:

```
CREATE TABLE ... ENCODING 'UTF8'
  -> SQLSTATE[42000]: Syntax error ... near '"UTF8"' at line 8
same statement with COLLATE = utf8mb4_unicode_ci
  -> created OK
```

Consequence: a fresh installation ends up with **47 of the 49 tables**, missing
`adm_forum_topics` and `adm_forum_posts`. Any use of the forum module then dies in
`Entity::__construct()`, because `Database::getTableColumns()` returns false and
`Entity::setColumnsInfos()` calls `fetchAll()` on it:

Image

```
Error: Call to a member function fetchAll() on false
  Database.php:619 -> Entity.php:947 -> Entity.php:155 -> Post.php:39 -> Topic.php:42
```

**Only fresh installations are affected.** `install/db_scripts/update_5_0.xml` steps 620 and 640
create the same two tables with a valid `COLLATE` clause, so anything upgraded from 4.x has them.
That is presumably why this has not been noticed.

**Open question: @Fasse ** whether an update step is needed for 5.x installations that were installed fresh with
the broken file and are therefore missing the two tables. `install/db_scripts/update_5_0.xml`
steps 620 and 640 only run for installations that come from 4.x.

Found through: the forum module tests of the new regression test suite, which were blocked by it.

kainhofer:
Both definitions ended with the PostgreSQL clause ENCODING 'UTF8', which MySQL and MariaDB reject, so a fresh installation was missing adm_forum_topics and adm_forum_posts and the forum module could not be opened. PostgreSQL never showed it because the query conversion cuts the table options.

**Open question:** @Fasse whether an update step is needed for 5.x installations that were installed fresh with
the broken file and are therefore missing the two tables. install/db_scripts/update_5_0.xml
steps 620 and 640 only run for installations that come from 4.x.

Fixes #2120

### DIFF
--- a/install/db_scripts/db.sql
+++ b/install/db_scripts/db.sql
@@ -263,10 +263,10 @@ CREATE TABLE %PREFIX%_forum_topics
     fot_usr_id_create           integer unsigned,
     fot_timestamp_create        timestamp           NOT NULL    DEFAULT CURRENT_TIMESTAMP,
     PRIMARY KEY (fot_id)
-    )
-    ENGINE = InnoDB
-    DEFAULT CHARSET = utf8mb4
-    ENCODING 'UTF8';
+)
+ENGINE = InnoDB
+DEFAULT CHARSET = utf8mb4
+COLLATE = utf8mb4_unicode_ci;
 
 CREATE UNIQUE INDEX %PREFIX%_idx_fot_uuid ON %PREFIX%_forum_topics (fot_uuid);
 
@@ -284,10 +284,10 @@ CREATE TABLE %PREFIX%_forum_posts
     fop_usr_id_change           integer unsigned,
     fop_timestamp_change        timestamp           NULL        DEFAULT NULL,
     PRIMARY KEY (fop_id)
-    )
-    ENGINE = InnoDB
-    DEFAULT CHARSET = utf8mb4
-    ENCODING 'UTF8';
+)
+ENGINE = InnoDB
+DEFAULT CHARSET = utf8mb4
+COLLATE = utf8mb4_unicode_ci;
 
 CREATE UNIQUE INDEX %PREFIX%_idx_fop_uuid ON %PREFIX%_forum_posts (fop_uuid);
 


Fasse:
Since this problem only exists within version 5.1 and not 5.0, you can add a new update step within 5.1 only for postgreSQL and add an `IF NOT EXISTS` to the `CREATE TABLE` command:

`CREATE TABLE IF NOT EXISTS %PREFIX%_forum_topics
(
    fot_id                      integer unsigned    NOT NULL    AUTO_INCREMENT,
    fot_uuid                    varchar(36)         NOT NULL,
    fot_cat_id                  integer unsigned    NOT NULL,
    fot_fop_id_first_post       integer unsigned,
    fot_title                   varchar(255)        NOT NULL,
    fot_views                   integer unsigned    NOT NULL    DEFAULT 0,
    fot_usr_id_create           integer unsigned,
    fot_timestamp_create        timestamp           NOT NULL    DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (fot_id)
    )
    ENGINE = InnoDB
    DEFAULT CHARSET = utf8mb4
    ENCODING 'UTF8';`

Fasse (changes requested):
View my last comment

kainhofer:
Thanks for the pointer. You are right, the wrong pgsql syntax was introduced at the end of February in the 5.1 branch. 

I have updated the PR to add the tables if they don't exist. HOWEVER: While 
`CREATE TABLE IF NOT EXISTS`
works on mysql and mariadb, the index and foreign key generation, i.e.
```
CREATE UNIQUE INDEX IF NOT EXISTS
 ADD CONSTRAINT %PREFIX%_fk_fot_cat         FOREIGN KEY IF NOT EXISTS
```
are mariadb-only syntax (starting in mariadb 10.x) and don't exist in mysql. 

1. Do we have any way to distinguish mariadb and mysql in the update*.xml?
2. Since this was introduced only during the development phase of 5.1, do we really need an update script for mysql/mariadb at all?

kainhofer:
@Fasse It's your decision whether we still need an update script for manual 5.1 installations that were not updated from 5.0. I have no idea how the index and keys should be handled in mysql (as opposed to mariadb), though, short of writing a PHP function to check before adding....

Fasse:
No we can also fix the script. Normally no one should use that version productive. But I don'T know what you mean with the different Index handlig between mysql and mariaDB. I think all syntax is also mariaDB kompatible.

So in my opinion we do not need update steps.

kainhofer:
Okay, so I'll skip the update_5_1.xml script to insert the missing table for new installations.

What I meant: MySQL and MariaDB have drifted apart in their SQL capabilities. Both support CREATE TABLE IF NOT EXISTS. However, we also need a unique index and a foreign key. 

CREATE INDEX IF NOT EXISTS:

- MySQL: https://dev.mysql.com/doc/refman/9.7/en/create-index.html -- does NOT support "CREATE INDEX IF NOT EXISTS"
- MariaDB: https://mariadb.com/docs/server/reference/sql-statements/data-definition/create/create-index - Supports "CREATE INDEX IF NOT EXISTS"


ALTER TABLE ... ADD FOREIGN KEY IF NOT EXISTS:

- MySQL: https://dev.mysql.com/doc/refman/9.7/en/alter-table.html -- Does NOT support "FOREIGN KEY IF NOT EXISTS"
- MariaDB: https://mariadb.com/docs/server/reference/sql-statements/data-definition/alter/alter-table -- Supports "FOREIGN KEY IF NOT EXISTS"


PostgreSQL is not an issue, since there the forum tables were always created properly.

PR #2126, opened 2026-08-24
Add comprehensive regression test suite for Admidio Core

### LINKED ISSUE #13: Testing Framework

kainhofer:
## Summary

This PR adds a comprehensive regression test suite for Admidio Core and integrates it into the Composer workflow through:

```bash
composer test:all
```

Documentation is available in our DokuWiki: https://www.admidio.org/dokuwiki/doku.php?id=en:entwickler:regression_test_suite

The main goal is not only to increase test coverage, but to establish a reliable regression boundary around actual Admidio behavior.

The suite therefore deliberately exercises real Admidio production code wherever possible:

* Entities and Services
* database writes and subsequent independent reads
* permissions and organization boundaries
* the production installer
* real CLI subprocesses
* filesystem operations
* image handling
* import/export
* SMTP delivery through Mailpit

Fixtures are only used to establish prerequisites. They do not reproduce the functionality that the test claims to verify.

## Motivation

> A regression test should invoke the real Admidio production path and independently verify its observable result.

For persistent functionality this normally means:

1. arrange the minimum required fixture data,
2. invoke the production Entity, Service or CLI command,
3. independently read the result back from the database, filesystem, another CLI process or an external test endpoint,
4. verify cleanup and isolation.

## Test architecture

The suite now contains several complementary layers.

### Unit tests

Fast tests for production logic that does not require a database, filesystem or network connection.

Unit tests do not initialize the regression database.

### Database and Service integration tests

Integration tests exercise real Admidio Entities and Services against a real database.

Covered areas include, among others:

* users and memberships
* roles
* organizations
* categories
* announcements
* menu entries
* profile fields
* inventory
* registrations
* OIDC/SSO configuration
* messages
* permissions and organization isolation
* database abstraction behavior

Service-path tests call the actual production Service and then independently verify what was persisted.

### CLI regression tests

The CLI is used as an acceptance boundary.

In addition to command contract tests, the suite contains real subprocess workflows. Each invocation runs the actual Admidio CLI bootstrap with its own database connection.

Mutating workflows cover, among others:

* users
* groups and memberships
* menu entries
* announcements
* web links
* inventory
* profile fields
* OIDC clients
* user relations
* private messages

Where applicable, one process performs a write and another process verifies the committed result.

This is intentionally stronger than calling the CLI callback inside the PHPUnit process.

### Filesystem integration tests

Filesystem tests exercise the real production Services for:

* documents
* photos and albums
* image conversion and thumbnails
* retained originals
* ZIP archive generation
* inventory CSV import/export

Tests verify both database state and the actual generated files.

All destructive filesystem activity is restricted to:

```text
tests/adm_my_files
```

A dedicated marker file must be present before cleanup is permitted.

### Mailpit integration

The mail regression test exercises the real Admidio mail path:

```text
PreferencesService
    -> Admidio Email
    -> PHPMailer
    -> SMTP
    -> Mailpit
```

The test then queries the Mailpit HTTP API and verifies that the message was actually delivered to a unique test recipient.

It deliberately does not depend on Docker's Mailpit health status, since the current container may incorrectly be reported as unhealthy even though SMTP and the HTTP API are operational.

## Test database setup

Database-backed test runs create a clean schema through the current production Admidio installer.

This avoids relying on a stale SQL fixture and also means the regression setup checks that the current production installation path can still create a usable database.

The test environment includes destructive-operation safeguards. In particular, the configured database must clearly be a dedicated test database.

A production or valuable development database must never be used for this suite.

Example:

```text
admidio_test
```

Database integration tests normally use transaction rollback for isolation.

CLI subprocess tests are a special case because subprocesses use independent database connections. Mutating CLI scenarios therefore explicitly clean up committed data.

## Administrator test context

Tests for functionality requiring administrative permissions do not bypass Admidio's rights system.

Instead, the suite loads the real administrator account and administrator membership created by the production installer and executes the corresponding Services with that normal Admidio user context.

This ensures that production permission checks remain part of the regression path.

## Docker test environment

The supplied Docker setup can be used to provide the required external services, including database servers and Mailpit.

Test configuration is provided through `.env.test`.

A corresponding `.env.test.example` documents the supported settings, including the Mailpit SMTP and API endpoints.

Environment variables may override the file configuration, which allows the same suite to be used in CI.

## Production issues found while implementing the suite

Building the regression suite already exposed several real issues in production code.

These fixes are included in this PR because they are required for the tested production paths to work correctly.

### Windows document path handling

The document folder containment validation compared normalized application paths with native `realpath()` results.

On Windows this caused valid child paths using `\` to fail a comparison against `/`.

The paths are now normalized before performing the existing containment/security check.

### OIDC user ID field persistence

A new OIDC client initialized `ocl_userid_field` directly in the Entity's internal column array.

That bypassed normal Entity change tracking, allowing the database default `usr_id` to be stored even when the Admidio default was intended to be `usr_login_name`.

Initialization now goes through the normal Entity value handling so the value is persisted correctly.

### Private-message deletion through the CLI

`Message::delete()` already implements the expected two-participant private-message deletion lifecycle, but the CLI command was restricted at the component-administration level before recipients could reach that logic.

Private-message recipients can now invoke the appropriate deletion path while stored email messages remain restricted to their sender.

### Headless inventory export

`inventory:export` is exposed as a CLI operation, but `ExportService` initialized the normal web `InventoryPresenter`, which in turn required web-only theme/session bootstrap state such as `THEME_URL`.

Inventory export now supports data-only presenter initialization so that the Service can also operate correctly from the CLI and other headless contexts.

## Regression-suite hardening

The implementation also addresses several sources of possible false-positive tests.

Among other things:

* fixtures use real Admidio Entities rather than in-memory substitutes;
* privileged tests use the real installed administrator instead of bypassing rights checks;
* Service tests verify persisted state independently;
* CLI mutation tests use real independent subprocesses;
* subprocess tests verify committed state;
* organization-boundary tests exercise production paths rather than reproducing a safe SQL query in the test;
* filesystem cleanup is protected by a dedicated test-directory marker;
* mail tests verify actual delivery rather than only checking whether a port is open;
* the current production installer is used instead of accepting an arbitrary existing schema;
* weak or placeholder assertions were removed or strengthened;
* PHP 8.4 compatibility warnings in test fixtures were addressed.

## Running the suite

Install the development dependencies:

```bash
composer install
```

Copy and configure the test environment:

```bash
cp .env.test.example .env.test
```

Then run:

```bash
composer test:all
```

Individual tests can also be run directly through PHPUnit, for example:

```bash
vendor/bin/phpunit tests/Integration/Services/InventoryServicePathTest.php
```

or:

```bash
vendor/bin/phpunit --filter testName
```

## Current result

After applying the complete implementation and the production issues discovered by it:

```bash
composer test:all
```

runs successfully on the tested development environment.

## Intentionally not covered yet

A few areas are deliberately left for follow-up work because the required production boundaries are not yet available or are still being developed:

* event CLI mutation workflows
* room CLI mutation workflows
* updater/migration lifecycle scenarios
* user import through the CLI

User import in particular should first have a suitable Service separation rather than adding a regression test that duplicates module logic.

Historic version-to-version database migration testing is also a separate lifecycle-testing step and should not be inferred from the current fresh-install coverage.

## Documentation

Developer documentation for setting up, running and extending the regression suite has been prepared for the Admidio developer wiki.

It describes:

* environment setup
* Docker usage
* database safety
* test layers
* fixture design
* Service-path testing
* CLI subprocess testing
* filesystem tests
* Mailpit
* adding regression tests
* common anti-patterns and troubleshooting

## Review notes

When reviewing newly added regression tests, the most important question is not simply whether the test passes.

The key question is:

> Would this test fail if the corresponding production Admidio behavior were broken?

A test that reproduces the expected behavior itself, only inspects fixture state, or bypasses the relevant production permission/database/filesystem path should not be considered sufficient regression coverage.

## Checklist

* [x] Regression suite can be run with `composer test:all`
* [x] Production installer is used for the test database
* [x] Database-backed tests use real Admidio APIs
* [x] Service mutations are independently verified
* [x] Administrative tests use real Admidio permissions
* [x] Mutating CLI tests run through independent subprocesses
* [x] CLI writes are verified after commit
* [x] Filesystem tests use a protected dedicated directory
* [x] Mail is sent through the real Admidio SMTP stack
* [x] Mail delivery is verified through Mailpit
* [x] Test environment configuration is documented
* [x] `composer test:all` passes

Fixes #13


kainhofer:
BTW, the test:integration and test:cli tests automatically run in github whenever a PR agains master is submitted. This PR is already an example, as the checks above show.... Mysql is skipped on purpose, but mariadb and pgsql were successful.

Fasse (approved):

PR #2128, opened 2026-08-26
PR: Preferences: generalize preference definitions; CLI hardening and fixes

kainhofer:
## Summary

The centerpiece of this PR is `PreferenceDefinitions`, a new class ([src/Preferences/Service/PreferenceDefinitions.php](src/Preferences/Service/PreferenceDefinitions.php)) that becomes the single source of truth for every core organization preference: its default value, scalar type, range/length, enumerated values, requiredness, sensitivity and semantic validation rule. Previously this knowledge was duplicated and partially hardcoded across `install/db_scripts/preferences.php`, `PreferencesPresenter`, the changelog masking list and ad-hoc CLI checks. Installer, web preferences form, changelog masking and `config:*` CLI commands now all validate and redact through the same definitions.

Alongside that, this PR carries a batch of CLI hardening/correctness fixes and a few unrelated fixes/refactors (contacts import/deletion, message delivery ordering, photo numbering) that came out of the same review pass.

## Preferences: one canonical definition per preference

- **`PreferenceDefinitions::table()`** describes every core preference as `default`/`defaultProvider`, `type` (`string`, `bool`, `int`, `enum`, `reference`), `values`, `minimum`/`maximum`/`maxLength`, `required`, `sensitive`, `internal`, and a `validator` (one of a fixed set of semantic rules, or a callable for cases the class itself can't express — e.g. cross-field checks). `coverageProblems()` rejects unknown keys so a typo in a definition can't silently disable a constraint.
- **`PreferenceDefinitions::register()`** lets a module or a plugin add its own preference the same way core preferences are declared; `validator`, `values` and `defaultProvider` accept a callable there too. `PluginAbstract` now builds its `defaultConfig` preferences through this instead of a private, ad-hoc default list.
- **`PreferencesService::seedDefaults()` / `removePreferences()`** insert or remove a registered preference's row in *every* organization. Previously a plugin's preferences were written only into the settings manager of whichever organization the installing administrator happened to be in, so every other organization silently had none. Installer, update, `organization:add` and plugin install/uninstall now go through this instead of hand-seeding from `preferences.php`.
- **Runtime fallback removed:** since a registered preference is guaranteed a real row in every organization, `SettingsManager::registerDefaults()` and its static default list are gone entirely — a missing preference row is an error again instead of being silently papered over at read time.
- `config:list`, `config:get`, `config:export` and `config:schema` (new) validate/redact against this same registry, so CLI and web form can no longer drift out of sync on what a preference accepts.

## CLI

- **`maintenance:run`**: one command running the recurring jobs (category/document repair, plus `session:cleanup`, `autologin:cleanup`, `sso:token-cleanup`) individually-reported, single non-zero exit if any fail — one crontab entry instead of four.
- **`--format=json-api`**: a versioned response envelope (schema version, stable error codes, structured warnings) alongside the existing plain `--format=json`, declared per command and checked by `cli:selfcheck`.
- **`config:schema`**: publishes supported preference names, types, ranges, enumerations and sensitivity for automation.
- Output-file safety: `--output` no longer follows a symlink or silently replaces an existing file (`file_put_contents`/`copy`/`rename` → explicit `--overwrite` semantics, secret exports created `0600` before their first byte). Applied consistently to `config:export`, the json-api envelope, and other file-producing commands.
- `--dry-run` for imports that can preflight, `--width`/`--no-truncate` for text tables, and a deprecation warning when a secret is passed as a CLI argument instead of via stdin.
- Permission/scoping corrections: `user:show` now applies profile field visibility (not just the coarser "may open a profile" check) to fields, memberships and relations; `relation:list`/`relation:delete` are scoped to the selected organization instead of operating globally; `profile:fields`/`profile:field-show`/`profile:options` only reveal internal configuration to a users administrator.
- New commands: `whoami`, `user:lock`/`user:unlock`, `group:permission-holders`.
- Correctness fixes: text output no longer requires `ext-mbstring`; CSV output no longer triggers the PHP 8.4 `fputcsv()` escape-parameter deprecation; a 2FA setup no longer prints the secret before the verification step can reject it; `user:list --former` now means "active and ended", not "no date filter at all"; an event participation is no longer counted as contact membership (matching `contacts_data.php`); the changelog-permission commands now use the same `ChangelogService` calls as the web changelog instead of a nonexistent method; `message:reply --attachment` (a rejected, non-existent option) no longer masks the real validation.
- Removed `organization:update --parent` — it let any organization become the parent without rejecting itself or its own descendants, and `OrganizationService` doesn't support the field; reparenting needs a real operation of its own.
- A global option the current command doesn't declare (e.g. an unsupported `--format`) is now rejected instead of silently parsed away; a repeated `--host`/`--organization` no longer bootstraps one value while the parser keeps another.

## Other fixes

- **Contacts**: import (spreadsheet reading, duplicate handling, login data, membership assignment, encodings/delimiters) extracted into `ContactImportService`, shared by the web wizard and the new `user:import` CLI commands. Ending/deleting a contact extracted into `ContactService`, replacing the divergent subset of rules the CLI previously reimplemented.
- **Messages**: the message is now written inside a transaction that only commits once the mail has actually been handed to the mailer, instead of persisting after send — a failure between delivery and save previously left no history and could cause a duplicate send on retry.
- **Photos**: the album row is locked while the next photo number is allocated (previously read from in-memory `pho_quantity` without locking), and a failed compaction now restores moved files on rollback instead of leaving the numbering decremented.


Fasse (changes requested):
I like the changes that we define more information to a preference. 

So we should do it more consistent. In my opinion every preference should have a type, so we should introduce string also as a type. enum should not be a type it should be another property of the preference because also an enum could be int or string or boolean.

Fasse:
Some more thoughts about the redesign of the preferences. We should also store these information in the database, but if we do it, we need to separate the values from the preference definitions so we need two tables, a preference table and the preference value table with this concept it would also be possible to allow user depending preferences. So the preference value table should have a column with the user ID. 

What do you think about it?

PR #2137, opened 2026-09-03
form.separator.tpl: Generalize to make the separator optional and add open/close caret support

kainhofer:
* Move underline from explicit <u> html tag to css
* Add `'separator_line' => false` option to disable the separator line (which also causes the label to be larger and underlined). This allows the form.separator.tpl to be used for "advanced settings" subareas instead of full top-level headlines.
* Add options` 'collapse' => 'html_div_id_to_be_collapses'` and `'collapsed' => true` to give the HTML ID of a div that should be shown/hidden with a open/close caret before the label text

Example:
In the PreferencesPresenter:
```php
        $formSSO->addSeparator(
            'sso_saml_advanced_settings',
            $gL10n->get('SYS_SSO_ADVANCED_PROPERTIES'),
            array('collapse' => "sso_saml_advanced", 'collapsed' => true, 'separator_line' => false)
        );

```
In the correpsonging preferences.*.tpl:
```smarty
    {include 'sys-template-parts/form.separator.tpl' data=$elements['sso_saml_advanced_settings']}
    <div id="sso_saml_advanced" {if !empty($elements['sso_saml_advanced_settings'].collapsed)} style="display: none;" {/if}>
        {include 'sys-template-parts/form.select.tpl' data=$elements['sso_saml_signing_key']}
        {include 'sys-template-parts/form.select.tpl' data=$elements['sso_saml_encryption_key']}
        {include 'sys-template-parts/form.checkbox.tpl' data=$elements['sso_saml_want_requests_signed']}
    </div>
```

image


Fasse:
@kainhofer Please fix the merge conflicts, after that this could be merged.

PR #2145, opened 2026-09-12
Move the member-sharing setting from preferences to organization db table

Fasse:
## Summary

Move the suborganization member-sharing configuration from the organization preferences table to the organization record itself.

## Changes

- Added `org_suborg_use_same_members` to `adm_organizations`.
- Updated the organization edit form, persistence logic, member-sharing scope, CLI output/options, and changelog mapping to use the new column.
- Removed the obsolete `contacts_suborganization_use_same_members` preference and its validation.
- Added an upgrade migration that transfers existing preference values and removes the old preference rows.
- Preserved legacy behavior for existing parent-child organization relationships during upgrades.
- Restrict enabling member sharing to top-level parent organizations only; child and standalone organizations cannot set the field to `true`.
- Added integration coverage for the column default and the parent/child validation rules.


kainhofer (approved):
Looks good, the settings belongs to the org entity, so the code becomes simpler. AFAICS, no change in behavior or UI for the user. 

- Only difference in behavior is that with a hierarchy of 3 or more organizations, inheriting is now only possible from top-level org. Inheriting from 2nd level org is no longer possible and its setting is not migrated. However, I'm not sure how many admidio installations have such a hierarchy... Probably not a real-life issue at all..
- Nitpick: db.sql: column type is shifted one space to the right compared to all other columns.
- CLI: setting --share-members when the org is not a top-level org doesn't show any error and prints a success, even though setValue() returns false. Again, I'm not sure whether nested orgs are really a target?

kainhofer (changes requested):
I tried the patch with current master, and it seems there is a 
<code>use Admidio\Preferences\Service\PreferenceDefinitions;</code>
missing in OrganizationService.php... Saving fails with unknown Admidio\Organizations\Service\PreferenceDefinitions....

Fasse:
Which saving failed? For me the saving of the organization form works. I could not reproduce that error.

Fasse:
@kainhofer Could you please describe the regression more detailed. I could not reproduce it. Otherwise I will merge this Pull Request.

PR #2148, opened 2026-09-12
Full Plugin system (install/enable/disable/update/uninstall; plugins providing modules or simple tweaks, etc.)

### LINKED ISSUE #851: Admin view as role "z"
A problem I am often facing is that users tell me that they are not able to do "x" or "y". Therefore an option would be nice, where I could see Admidio like Role "z". Now i have to keep a test-account and ad me to certain roles and check, if an option is available.

### LINKED ISSUE #1969: disable functionality in plugin manager
Once there is a public plugin system in place with plugins that are not distributed together with the core, you always have to assume that there are plugins that will break the whole system (e.g. outdated code, bad coding, etc.). In that case, the plugin system needs to disable the plugin to allow admidio to function.

_Originally posted by @kainhofer in https://github.com/Admidio/admidio/issues/1964#issuecomment-3865512102_

kainhofer:
## Summary

This significantly extends, generalizes and simplifies Admidio's plugin manager: a plugin is now 
simply a directory below `plugins/` with a JSON manifest and a plain PHP entry file. No base class
is required, no `PluginAbstract`, no scanning the filesystem for conventionally-named classes. 
Everything a plugin needs is either declared in the manifest, registered through the hook system 
of #2127, or accessible as a module endpoint: autoloading, settings that become ordinary 
Admidio preferences, a generated settings form, a widget on the overview page, a preferences 
panel, its own CLI commands, its own test suite. Module pages provided by plugins can be linked
directly into the /modules/ directory, so modules provided by plugins are not second-class citizens.

The plugin manager now also supports installing and updating plugins, enabling/disabling installed 
plugins, and removing plugins. Plugins are also no longer restricted to Overview widgets, but 
provide arbitrary functionality to Admidio.

The previous eight built-in Overview plugins (`birthday`, `calendar`, `announcement-list`, `event-list`,
`latest-documents-files`, `login-form`, `random-photo`, `who-is-online`) are converted to it, and an
update step migrates existing installations. 
The PR also provides a sample plugin to be used as a showcase and a starting point for custom 
plugin development: `plugins/hello-world/` is a new, fully commented example that shows every 
convention in one place; its own `README.md` provides a short-form guide to plugin development.

## Why this matters: a real third-party plugin

The clearest evidence this extended, yet simpler plugin system enough for a real extension is
**[admidio-impersonate](https://github.com/kainhofer/admidio-impersonate)** — a third-party plugin,
developed entirely outside this tree against the manifest and hook API this PR introduces, that lets
an administrator act as another user to see what they see. It is not a toy:

* its settings — the impersonation timeout, who may be impersonated — are declared in its manifest
  and get a generated preferences form for free;
* it adds an *Act as …* icon to every row of the contacts list through an ordinary hook, with no core
  code aware that it exists;
* it keeps its own database table (`db_scripts/install.sql` / `uninstall.sql`) and its own migration
  for future versions;
* it marks every change made while impersonating with an origin comment through the existing
  `LogChanges::setOriginComment()`, and the change history shows that comment as a tooltip;
* it ends the impersonation on logout instead of the session, and returns the visitor to the
  homepage of its own choosing afterward — both through the two small `logout.php` hooks below, with
  no plugin-specific code in `system/logout.php`;
* it ships its own PHPUnit suite, picked up by `composer test:*` through `plugins/*/tests/<suite>`,
  exactly like a first-party plugin;
* it is packaged and installable exactly the way this PR intends: `admidio plugin:archive`, upload,
  or a plugin store entry.

If a plugin this involved (modifying the user that works with Admidio on the fly, etc.) needs 
almost nothing from the core beyond what is in this PR, the API is in the right place.

## What is in this PR

* **Manifest** (`plugin.json`): name, version, `requires` (Admidio/PHP version constraints, PHP
  extensions, other plugins, and — advisory only — Admidio modules), `autoload` (PSR-4, may not
  shadow `Admidio\`), `settings` (typed preferences with generated form fields), `preferences`
  (which tab and position its panel gets).
* **`PluginRegistry`**: the single place that answers which plugins exist and in which of six states
  (`available`, `disabled`, `enabled`, `update`, `broken`, `orphaned`) they are. Reading a plugin
  never executes its code.
* **`PluginLoader`**: loads the plugins enabled for a request — autoload, language files, registered
  settings, the entry file — and reports a failure instead of taking the request down with it.
* **`PluginInstaller`**: install / update / enable / disable / remove, four operations with
  deliberately different scope (installation-wide vs. per-organization) and one destructive one.
* **`PluginWidget`** and **`PluginPanel`**: the overview widget and the preferences panel are now
  ordinary hooks instead of files the core scans for by name or content.
* **`PluginPages`**: a plugin page is always reachable below `plugins/<id>/modules/`, and optionally
  published under the ordinary `modules/<id>/` URL when an administrator turns that on — never both
  hardcoded.
* **`PluginStore`** and **`PluginPackage`**: a JSON catalogue for installing/updating from a published
  source, and a checked ZIP archive format for upload and distribution (size/entry limits, a two-pass
  gate before anything is written to disk).
* **`admidio plugin:*`**: the full lifecycle from the command line — `list`, `show`, `install`,
  `inspect`, `enable`, `disable`, `update`, `archive`, `remove`, `store-list`, `store-show`,
  `store-refresh` — plus `plugins/<id>/cli.php` for a plugin to register commands of its own.
* **`system/logout.php`**: the `logout_target` resolver and the `logout_homepage` filter described
  above.
* **`modules/changelog/changelog_data.php`**: shows the `log_comment` of a change as a tooltip on its
  row number, so a change made in a marked context — the command line, or a plugin like
  admidio-impersonate — can be told apart from an ordinary one. Neither addition is plugin-specific.
* Tests: unit, integration and CLI coverage of the registry, loader, installer, store and package,
  plus `plugins/hello-world/tests/` as a worked example of testing a plugin from where it lives.

65 commits, 166 files, roughly +14.3k/-4.8k lines against current `master`.

## Documentation

The full developer reference — manifest schema, every hook a plugin can use, the lifecycle, the
archive format, and a table of what changed for a plugin written against the previous (5.0) plugin
manager — is attached below as an updated DokuWiki page for
`en:entwickler:plugin_manager`. The design rationale and the alternatives that were considered are
in the attached concept document.

  * New Plugin development tutorial: https://www.admidio.org/dokuwiki/doku.php?id=en:entwickler:plugin_development
  * The Dokuwiki page en:entwickler:plugin_manager needs updating. My suggested text for it is attached: [2026-09-06_plugin_manager_dokuwiki.txt](https://github.com/user-attachments/files/32149625/2026-09-06_plugin_manager_dokuwiki.txt)

Design and technical concepts (written by ChatGPT and Claude):
  * Original concept by ChatGPT:  [2026-08-04_admidio_generic_plugin_system_concept.docx](https://github.com/user-attachments/files/32149675/2026-08-04_admidio_generic_plugin_system_concept.docx)
  * Extended concept by Claude: [2026-08-26_admidio_plugin_system_full_concept.md](https://github.com/user-attachments/files/32149701/2026-08-26_admidio_plugin_system_full_concept.md)
  * Claude review of ChatGPT's concept and consolidated version: [2026-08-28_admidio_plugin_system_consolidated_concept.md](https://github.com/user-attachments/files/32149726/2026-08-28_admidio_plugin_system_consolidated_concept.md)

What is yet to be implemented: a **PLUGIN STORE**
  * Plugin store that provides all available plugins with information in JSON format. 
  * Concept: [2026-08-30_admidio_plugin_store_concept.md](https://github.com/user-attachments/files/32149756/2026-08-30_admidio_plugin_store_concept.md)

## Notes for reviewers

* This branch is rebased onto current `master` and therefore already contains #2127 (the hook
  system) and the merged SSO work; the diff above is Plugin System's own commits only.
* The commit history is intentionally granular — each commit is one reviewable step of the
  conversion (one built-in plugin at a time, one manifest capability at a time) rather than a single
  squash, and each commit message explains the "why" of that step.
* `plugins/hello-world/` is safe to install on a running site to see the runtime work end to end
  without touching anything the site already depends on.
* To see screenshots of the plugin manager in action, see the impersonate plugin: https://github.com/kainhofer/admidio-impersonate

This PR fixes #1969; the impersonate plugin (relying on the functionality provided by this PR) fixes #851.


src/Infrastructure/Plugins/Plugin.php
```diff
@@ -0,0 +1,809 @@
...
+    {
+        global $gSettingsManager;
+
+        if (!isset($gSettingsManager)) {
+            return array();
+        }
+
+        $disabled = array();
+
+        foreach ($this->requires['modules'] as $module) {
```
MightyMCoder, line 522: Image

We should check whether `$this->requires['modules']` is set before iterating over it. Alternatively we could initialize it in general with an empty array when no manifest file is found (line 188):
```php
        $empty = array('admidio' => '', 'php' => '', 'extensions' => array(), 'plugins' => array(), 'modules' => array());
```

src/UI/Presenter/PluginsPresenter.php
```diff
@@ -41,244 +62,632 @@ public function createList(): void
...
+    {
+        if ($plugin === null) {
+            return array();
+        }
+
+        $diagnostics = array();
+        if ($plugin->error !== null) {
+            $diagnostics[] = $plugin->error;
+        }
+        $diagnostics = array_merge($diagnostics, $plugin->checkRequirements(PluginRegistry::getEnabledVersions()));
```
MightyMCoder, line 518: If a plugin is missing its manifest file this will report the same error twice:
Image

Wouldn't it be enough to leave the diagnostics array uninitialized (so without `if ($plugin->error !== null) {
            $diagnostics[] = $plugin->error;
        }`) and just use the error returned by `$plugin->checkRequirements(PluginRegistry::getEnabledVersions())`?

src/Infrastructure/Plugins/PluginInstaller.php
```diff
@@ -0,0 +1,431 @@
...
+     * @throws Exception
+     */
+    public static function disable(Plugin $plugin): void
+    {
+        global $gSettingsManager;
+
+        if (!PluginRegistry::isInstalled($plugin->id)) {
+            throw new Exception('SYS_PLUGIN_NOT_INSTALLED', array($plugin->id));
+        }
+
```
MightyMCoder, line 280: After disabling a plugin the menu should be refreshed. Otherwise the entries remain visible and clicking them results in an error message:

Image

Image

MightyMCoder (changes requested):
I’ve only had time for a quick review of this PR and have commented on the things that stood out to me. Deeper issues resulting from these changes will likely become apparent over time.

I haven’t tested the CLI functionality or run the tests.

Fasse (changes requested):
When there are "old" plugins that don't have the new structure, the entries in the Plugin-Managers shows errors.

Image

The version of @MightyMCoder list those plugins as not compatible with the current plugin manager. I think you should reimplement such a view to those plugins. They should not be labeled as faulty more something like not compabible to the plugin manager. @MightyMCoder Which naming do you have used?

MightyMCoder:
> When there are "old" plugins that don't have the new structure, the entries in the Plugin-Managers shows errors.
> 
> 
> 
> Image
> 
> 
> 
> The version of @MightyMCoder list those plugins as not compatible with the current plugin manager. I think you should reimplement such a view to those plugins. They should not be labeled as faulty more something like not compabible to the plugin manager. @MightyMCoder Which naming do you have used?


I used the string 'SYS_PLUGIN_NO_INTERFACE' which translates to 'The plugin has no interface.'

That wasn't ideal either and I agree with you @Fasse we should mark/label it as you suggested.

PR #2149, opened 2026-09-13
Group event settings into responsive tabs matching the profile layout

Fasse:
## Motivation

The event form previously displayed general, registration, and recurrence settings on one long page. This change groups related settings, reduces the number of options visible at once, and adopts the familiar layout of the profile page.

## Changes

- Organize the form into three tabs: **Basic data**, **Registration**, and **Recurrence**.
- Display the sections as an accordion on mobile, matching the profile page.
- Load all fields immediately, without additional requests or duplicated controls.
- Keep registration settings visible but disabled when registration is off, preserving their values.
- Keep the recurrence interval and end condition visible but disabled when recurrence is off. Show weekdays, occurrence count, and end date only when needed.
- Open the relevant section when a field fails browser validation.
- Keep the save button accessible outside the sections.
- Add dedicated translated headings, including “Anmeldung” and “Wiederholung” in German.
- Use the standard dark text color for active accordion headings and arrows throughout the simple theme.

## Validation

- PHP syntax check passed for `EventFormPresenter.php`.
- Updated language XML files parse successfully.
- `git diff --check` passed.
- Visual layout and browser interactions have not been independently verified.


kainhofer:
Don't the event start/end date and the recurrence setting belong together? After all, when you edit the recurrence, otherwise you only see that the event occurs e.g. every tuesday and thursday, but you don't see when... Similarly, when you only look at the basic data, you don't see when the event actually occurs, only when the first occurence happens...

Fasse:
Thanks for your quick feedback @kainhofer  Please have a look at the implementation now. I restore the reccurence settings in the main tab.

src/UI/Presenter/EventFormPresenter.php
MightyMCoder, line 436: Your changes look great!

If we also want to scroll to the currently opened accordion item, we could add something like this:
```js
// Jump to the opened accordion item on small screens, so the user does not have to scroll manually.
$(document).on("shown.bs.collapse", "#adm_events_edit_form .accordion-collapse", function() {
    var header = $(this).closest(".accordion-item").find(".accordion-header").first();

    if (header.length) {
        $("html, body").animate({
            scrollTop: header.offset().top
        }, 500);
    }
});
```
This would only align the user experience with the preferences page and has no functional impact.

MightyMCoder (approved):

Fasse:
Within this dialog we don't need to jump directly to an accordion entry. There is no functionality that some could start at a special tab / accordion entry.

PR #2154, opened 2026-09-13
Test PR for PR regression tests

kainhofer:
## THIS PR IS **NOT** MEANT TO BE MERGED!

This PR on purpose contains five broken commits, each of which breaks a particular part of Admidio. The purpose of this PR is to test whether the github regression test really catch those broken commits in the regression test suite.

1. Weblink save() silently rtrims the trailing / off lnk_url before writing | Weblink.php:171 -> Both DB types should fail
2. Category sequence query uses IFNULL(MAX(...), 0) instead of COUNT(*)  | Category.php:428 -> Should fail only on PostgreSQL
3. link:update passes $new = true to applyLinkOptions(), so it now demands --category | CoreTasks.php:6011  -> CLI test should fail
4. RolesRights::hasRight() treats "no roles assigned" as "open to everyone" | RolesRights.php:181 -> a real privilege escalation
5. Wrong operator in the fingerprint guard (|| → &&), so *_timestamp_change is never recorded | Entity.php:1386

The regression test checks done by github should NEVER be successful, with mysql and pgsql having different number of failures.


kainhofer:
@Fasse : Feel free to close this PR. It's only purpose is to test github's regression tests and to let you know/see that the regression tests on github actually work.

PR #2158, opened 2026-09-14
#2153 Fix: pin explicit collation on Date/Timestamp search conditions

### LINKED ISSUE #2153: Global contacts search fails with "Illegal mix of collations" when a Date/Timestamp column is in the list configuration and the database uses utf8mb4
### Environment
- Admidio version: 5.0.15
- Database: MariaDB (run `SELECT VERSION();` to fill in exact version)
- Database charset: utf8mb4 / utf8mb4_unicode_ci (fully converted from legacy utf8mb3/utf8)
- PHP version: (run `php -v` to fill in)

### Summary
The contacts list search (DataTables AJAX endpoint `modules/contacts/contacts_data.php`) fails
with a SQL error whenever the active list configuration includes a Date-type profile field, or
a Date/Timestamp special field (e.g. "Geändert am" → `usr_timestamp_change` /
`mem_timestamp_change`, or `mem_begin`/`mem_end`), on a database that has been converted to
utf8mb4.

### Error
SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (utf8mb4_unicode_ci,IMPLICIT)
and (utf8mb4_bin,NONE) for operation 'like'

FILE: src/Infrastructure/Database.php   LINE: 873   CALL: Database->showError()
FILE: modules/contacts/contacts_data.php   LINE: 306   CALL: Database->queryPrepared()

### Root cause
ListConfiguration::getSearchConditions() (src/Roles/Entity/ListConfiguration.php) builds a
search fragment for Date-type profile fields and for Date/Timestamp special fields like this:

    $arrSearchConditions[] = 'COALESCE(' . $listColumn->getValue('lsc_special_field') . ', \'1900-02-01\')';

MariaDB assigns the result of COALESCE(dateColumn, '1900-02-01') a utf8mb4_bin collation with
derivation NONE — this happens regardless of the actual column/connection charset (we
reproduced it both on a legacy utf8mb3 database and after fully converting the database to
utf8mb4). This result is then concatenated with the other, correctly utf8mb4_unicode_ci search
columns and compared via LOWER(CONCAT(...)) LIKE ... in contacts_data.php (~line 306), which
MariaDB rejects outright.

### Suggested fix
Pin an explicit collation on the two COALESCE expressions involving Date/Timestamp columns
(around lines 634 and 647 in our copy of the file):

    $arrSearchConditions[] = 'COALESCE(' . strtolower($gProfileFields->getPropertyById($lscUsfId, 'usf_name_intern')) . ', \'1900-02-01\') COLLATE utf8mb4_unicode_ci';

    $arrSearchConditions[] = 'COALESCE(' . $listColumn->getValue('lsc_special_field') . ', \'1900-02-01\') COLLATE utf8mb4_unicode_ci';

We applied this two-line patch locally and confirmed it resolves the issue with no observed
side effects.

### Additional note
src/Infrastructure/Database.php still opens the PDO connection with charset=utf8 (legacy
3-byte) and has a "// TODO: change to charset=utf8mb4 if we change charset in DB to utf8mb4"
comment, suggesting utf8mb4 databases aren't officially supported/tested yet. Might be worth
tracking full utf8mb4 support as a follow-up, since it's the current MySQL/MariaDB-recommended
default and needed for full Unicode (emoji etc.) support.

### Steps to reproduce
1. Have a database fully converted to utf8mb4 (all tables/columns utf8mb4_unicode_ci).
2. In a contacts list configuration, include any Date-type profile field, or a special field
   such as "Geändert am".
3. Use the global search box on the Kontakte page and enter any search term.
4. Observe the SQL error above.

olliengel:
Fixes #2153

### Summary
The contacts list search (DataTables AJAX endpoint modules/contacts/contacts_data.php) fails
with a SQL error whenever the active list configuration includes a Date-type profile field, or
a Date/Timestamp special field (e.g. "Geändert am" / usr_timestamp_change / mem_timestamp_change,
or mem_begin/mem_end), on a database that has been converted to utf8mb4.

### Root cause
ListConfiguration::getSearchConditions() builds a COALESCE(dateColumn, '1900-02-01') expression
for Date/Timestamp columns. MariaDB assigns this a utf8mb4_bin collation with derivation NONE,
which then clashes with the other, correctly utf8mb4_unicode_ci search columns when concatenated
and compared via LIKE.

### Fix
Pin an explicit COLLATE utf8mb4_unicode_ci on the two affected COALESCE expressions, so MariaDB
doesn't have to resolve an ambiguous collation. Tested and confirmed working on a production
database with the reported error.


Fasse (changes requested):
Since we also support Postgres you must do a database dependend SQL.

olliengel (commented):
Fixed in 5d0750b22 — the COLLATE utf8mb4_unicode_ci suffix is now only appended when running on MySQL/MariaDB. Added a $dateCollateSuffix variable, using the same DB_ENGINE === Database::PDO_ENGINE_PGSQL check that's already used elsewhere in this file (numeric column sort casting). PostgreSQL is unaffected since it doesn't hit this collation issue in the first place.

Fasse:
Hmm, for my environment this fix is not working:
Bildschirmfoto 2026-09-18 um 19 08 28

olliengel:
Good catch, thanks! The previous fix hardcoded COLLATE utf8mb4_unicode_ci, which breaks on installations whose tables are still on the legacy utf8/utf8mb3 charset — that's exactly the error you got. Root cause: Admidio's MySQL DSN always forces charset=utf8 in the connection (see the TODO in Database.php), independent of what charset the tables were actually created with, so this varies per installation/host.

Fixed in 255a8da7: instead of assuming a fixed charset, the collation is now looked up dynamically from an existing column (adm_users.usr_login_name via SHOW FULL COLUMNS) and reused — so it works whether the installation is on utf8/utf8mb3 or utf8mb4. PostgreSQL is still untouched. Could you retest in your environment when you get a chance?

Fasse:
Thanks for this fix. Now it works!

PR #2161, opened 2026-09-16
Cli permissions

### LINKED ISSUE #2136: Restrict CLI access
Currently, the new command line utility ./admidio allows the caller to perform any action on the admidio installation. Rationale is that a shell user with access to the admidio directory also has read access to the config file and thus all database credentials.

On the other hand, most non-reporting commands require the --as=username flag to indicate which user should be reported as actioning user. There is no authentication, though, simply giving the --as=username allows the caller to perform admidio actions as the given Admidio user.

Issues:
* Any shell user with read access to the Admidio directory can execute the CLI and perform actions. Some Admidio administrators might not even be aware of this.
* Shell users from other hosts on the same server with read-access to the admidio directory can modify admidio with the CLI (they have read access to the DB credentials, so no additional security issue, but it makes changing admidio easier)
* Having to give the --as parameter with basically any call is annoying.

Solution:
* Add a configuration flag to enable/disable CLI access at all
* Add a configuration field to restrict CLI access to certain shell users (username or system user ID)
* Let the admin select a default --as user for the CLI, overrideable by an explicit --as=username

kainhofer:
## Restrict CLI access (#2136)

The Admidio command line can currently be used by anyone with shell access to the
installation. This adds a configuration file, `adm_my_files/cli-config.php`, that 
* configures whether CLI may be used at all, 
* by which system accounts, 
* for which commands (a blacklist and a whitelist) and 
* as which Admidio user (i.e. restrict the --as users) and that 
* supplies defaults for options one otherwise repeats on every call. (e.g. supply a default for --as, so it isn't needed with every call any more)

**The command line is disabled unless the file exists and sets `$gCliEnabled = true`.** A new
installation therefore offers no CLI access until an administrator asks for it. 

Doing a free Installation from CLI is the only possible CLI use without the config file.

### Please read this before reviewing the restrictions

The cli-config.php file expresses a policy; it is not a security boundary, and the PR does not pretend
otherwise. Whoever can read `adm_my_files/config.php` holds the database credentials and
reaches the same data without the command line — and can copy the source tree, so no check
written in PHP can stop them. What the file buys is a safe default, an auditable statement of
intent, and a way to open the CLI to a cron account without handing it everything. It can be used 
to safeguard against inadvertent/wrong uses of the CLI. 

What *does* enforce it is the operating system, so `cli:selfcheck` now reports the relevant
file permissions, and the documentation names the measures that actually hold
(`chmod 750 adm_my_files`, or `sudoers` with `config.php` readable only by the Admidio user).

### The configuration file

```php
$gCliEnabled         = false;                       // master switch
$gCliAllowedUsers    = 'reinhold, %admidio-cli';    // name, uid, %group, %gid — empty = everyone
$gCliAllowedCommands = 'maintenance:*, user:list';  // fnmatch patterns, empty = all
$gCliDeniedCommands  = 'database:backup';           // checked first, wins
$gCliAllowedActors   = 'cronjob';                   // accounts --as may assume
$gCliLogFile         = 'logs/admidio-cli.log';      // relative = adm_my_files, empty = off
$gCliDefaults        = array('as' => 'reinhold', 'host' => 'www.example.org');
```

- **Groups** use the `sudoers` notation `%name` / `%gid` and count supplementary membership, so
  an operator is added with `usermod -aG admidio-cli` and the file stays unchanged.
- **Identity** comes from `posix_geteuid()`, never from `$USER`/`$USERNAME` — the environment is
  set by the caller. On Windows it is read from `%SystemRoot%\System32\whoami.exe` (called by
  absolute path), which yields the account name, its SID and its groups. If neither is
  available and a restriction is configured, the command is refused with a message that says
  the account could not be determined rather than that it is not on the list.
- **Option defaults** resolve as `--option` → `ADMIDIO_<OPTION>` → `$gCliDefaults` → built-in,
  per command as well (`'user:list' => array('limit' => 100)`). `--yes` and `--overwrite` can
  never be defaulted. `ADMIDIO_NO_DEFAULTS=1` runs a single call without the file's defaults,
  which is also how the regression tests stay independent of the developer's configuration.
- A refused command exits **5** (rejected) and reports which setting refused it.
- `help`, `list`, `version`, `completion` and `cli:defaultconfig` stay available while the CLI
  is disabled — otherwise nobody could find out how to enable it.

### Getting the file

- `admidio cli:defaultconfig` prints the documented template to stdout and writes nothing;
  `--output=adm_my_files/cli-config.php` creates it (refusing to clobber without `--overwrite`).
  It needs no database, so it still works when something is broken.
- New installations (browser and CLI) write the template in its disabled form.
- Existing installations get it from update step 1240 in `update_5_1.xml`; a data directory that
  cannot be written only produces a warning, never a failed update.
- Preferences → System overview shows the file, the state, the restrictions and the configured
  acting user resolved to a real name. Read-only: a switch that anything holding the database
  credentials could flip would restrict nobody.

### Notes for review

- No schema change, no new preference; the update step only writes a file.
- An existing installation that never creates the file loses CLI access on upgrade. That is the
  intent of the issue, and it is what the panel and the update-written template announce.
- Two commits: the command line itself, then the preferences panel.
- `tests/Cli/CliPolicyTest.php` covers matching, the cascade and every refusal in process; two
  subprocess tests cover the printed template and the environment cascade through the real
  binary. Suites on this branch: CLI 71, unit 166, integration 463, all green.
- Because the subprocess tests run the real `./admidio`, they now need the CLI enabled in the
  checkout's own `adm_my_files/cli-config.php`. When the checkout is not an installation (CI),
  the harness writes that file and removes it again; when it is, it refuses to touch it and
  says so. Documented in `tests/README.md`.

Closes #2136


Fasse (changes requested):
Why do you set the command line prefences at such a prominent place?
In my opinion this is a tool for a smaller audience so I would set it in last place of this setting list.

